PR #2, opened 2022-05-06
Make sdtab.T work with interfaces again, and stretch to terminal width

scott-cotton:
Using interface, we can avoid copying the struct at the call-site,
and more easily make adaptors for different types just by wrapping them.
At the definition site, it is slightly more verbose but at the call site
of AddRow, it is less verbose.

This PR however keeps the pure struct tag based approach.

Also, modified the calculation of table widths to stretch to terminal
width.


foxish:
Closing this, please reopen if this work is continued.

PR #4, opened 2022-05-18
Polish UX for sandbox commands

enisoc:
- [sandbox create: Add wait timeout and spinner.](https://github.com/signadot/cli/commit/ee0e8a2edbb58ab73e60cf127ed667eff20fa8a0)
- [sandbox delete: Wait for sandbox to fully terminate.](https://github.com/signadot/cli/commit/5142de031d5dc56bdbab9773de76d7a79fb2d4a9)
- [sandbox create: Print info on accessing the sandbox.](https://github.com/signadot/cli/commit/8bb0f363980e159f9ac05757e4f9c31869fed823)
- [sandbox get: Print details instead of a summary table.](https://github.com/signadot/cli/commit/ffca2480d1364fb750a62f7a2b2589c88e807daf)
- [Add 'sandbox get-status' command.](https://github.com/signadot/cli/pull/4/commits/d519a12d722acbe93db9fa560ec42658aa6aea94)
- [sandbox create/delete: Use new /status API.](https://github.com/signadot/cli/pull/4/commits/86dcdf22b01edb5f6cf5d81aa618b98aa53eee8e)


scott-cotton:
For delete, I am getting the attached image for the output, which is odd because it is 
repeating the same (truncated) message but still has the spinner.

Screenshot 2022-05-19 at 16 33 56

enisoc:
I was able to reproduce that too by making my terminal window narrower. So it seems that the spinner library ([yacspin](https://github.com/theckman/yacspin)) doesn't work properly when you try to print a message that doesn't fit in the terminal.

I don't see an issue for that library, but the other popular Go spinner library seems to have the [same issue](https://github.com/briandowns/spinner/issues/113). It appears to be a common challenge for spinners, though [potentially solvable](https://github.com/sindresorhus/ora/issues/74).

As a quick workaround, perhaps we could truncate the message to fit in the terminal while the spinner is running, and then print the full message only at the end when the spinner stops.

scott-cotton (commented):
Regarding the print package, it seems to me that it is starting to scatter the consolidation of a command implementation under internal/command/<command>

Would it be possible to have things like print.{ClusterTable, SandboxDetails} in its own file  and function under internal/command/<command>?  In this scenario I would consider what's remaining in print to be minor helper functionality such as making the json more like the yaml, and it seems it would retain the nice property of the current organisation in this PR of breaking out individual cases when they start to become more cumbersome in the switch statement.

scott-cotton:
> @scott-cotton I implemented the above workaround. PTAL.



> 

Looks like that will work, great.  At the same time, I'd personally have preferred being more conservative about the packages we import, as often enough the dependencies + such workarounds end up taking more time in the long run than building directly what we need, starting with minimal features.  A spinner seems like a it may be such a case.

enisoc:
> Regarding the print package, it seems to me that it is starting to scatter the consolidation of a command implementation under internal/command/

Good point. I moved the command-specific printers back into the same package as the rest of each command's implementation.

> A spinner seems like a it may be such a case.

I thought so too so I started without a spinner at all and just trying to rewrite the status line. I started running into terminal nuances and at some point I decided it would be a better use of my time to just pick an off-the-shelf library because I don't know enough about this terminal magic.

internal/command/sandbox/delete.go
```diff
@@ -79,7 +82,53 @@ func delete(cfg *config.SandboxDelete, out io.Writer, args []string) error {
...
+				return true
+			}
+
+			// Otherwise, keep retrying in case it's a transient error.
+			spin.Messagef("error: %v", err)
+			return false
+		}
+		status := result.Payload.Status
+		if status.Ready {
+			spin.Message("Waiting for sandbox to begin terminating.")
```
foxish, line 123: seems like a weird error message - maybe we group this under "Waiting for sandbox to terminate"?

foxish (approved):
LGTM aside from the one comment

internal/command/sandbox/delete.go
enisoc: Done.

PR #5, opened 2022-06-03
sandbox create: Support --output flag.

enisoc:
This adds support for the `--output` flag to the `sandbox create` command.

As part of this, I separated log output and sent it to stderr so that it won't get combined with the output JSON/YAML. I also updated other commands to use the same convention of sending log output to stderr for consistency.

### DIFF
--- a/internal/command/sandbox/create.go
+++ b/internal/command/sandbox/create.go
@@ -8,6 +8,7 @@ import (
 	"github.com/signadot/cli/internal/clio"
 	"github.com/signadot/cli/internal/config"
 	"github.com/signadot/cli/internal/poll"
+	"github.com/signadot/cli/internal/print"
 	"github.com/signadot/cli/internal/spinner"
 	"github.com/signadot/go-sdk/client/sandboxes"
 	"github.com/signadot/go-sdk/models"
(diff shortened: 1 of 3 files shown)


internal/command/sandbox/create.go
```diff
@@ -51,34 +52,42 @@ func create(cfg *config.SandboxCreate, out io.Writer) error {
...
+			fmt.Fprintf(log, "  signadot sandbox get-status %v\n\n", req.Name)
 			return err
 		}
 	}
 
-	// Print info on how to access the sandbox.
-	sbURL := cfg.SandboxDashboardURL(resp.SandboxID)
-	fmt.Fprintf(out, "\nDashboard page: %v\n\n", sbURL)
+	switch cfg.OutputFormat {
+	case config.OutputFormatDefault:
```
jmsktm, line 73: What value does `cfg.OutputFormat` get when:
a. When the flag is provided with just no value. Empty string?
b. When the `--output` flag is not provided. nil / empty?

Just trying to make sure it doesn't land in the default case if `nil` is a possibility causing it to error out.

I took a look at [this](https://github.com/signadot/cli/blob/424bbad106d97e684ab5625470ead9e5e0b784b4/internal/config/root.go#L29) and [this](https://github.com/signadot/cli/blob/4cbf27dc076dd4f3466fd5379a662011b15285a4/internal/config/flag_types.go#L18). But I couldn't be certain.

jmsktm (approved):
I've made a comment on the default value for the `--output` flag under two cases. Feel free to merge if it defaults to an empty string in both the cases.

internal/command/sandbox/create.go
jmsktm: To add to that, if it defaults to an empty string in absence of the `--output` flag, should we write to log in that case?

PR #6, opened 2022-07-18
Update to V2 API.

enisoc:


scott-cotton:
I've started testing it and so far it seems to work fine.

I am wondering if we should get rid of get-status since we are putting status in the get result.

scott-cotton:
Second item is perhaps we should add the list and delete cluster token functionality.

enisoc:
> I am wondering if we should get rid of get-status since we are putting status in the get result.

Yeah I debated that as well while working on this. At this point, the only benefit of `get-status` is that it might save the user from having to extract info with `jq` or `awk` if they want to check status from their own script. It seems like a small benefit though so I say let's remove it. Even if we still want to address that use case, we probably wouldn't do it with a separate command now that `get` returns status info.

> Second item is perhaps we should add the list and delete cluster token functionality.

I was thinking we'd just fix what's already there in the first pass, but this does seem quick so I'll add it if I get a chance.

scott-cotton:
I just accidentally put 'tags' at the top level of a sandbox json file instead of under 'spec' and it accepted the request without error and silently dropped the tags.  Should we check for that kind of thing?

enisoc:
@scott-cotton I've added cluster token list/delete. PTAL.

> I just accidentally put 'tags' at the top level of a sandbox json file instead of under 'spec' and it accepted the request without error and silently dropped the tags. Should we check for that kind of thing?

Do you mean like a client-side schema check to see if you're trying to set an unknown field? That seems reasonable, but we'd probably want to put some thought into it as a separate effort/feature.

scott-cotton:
> @scott-cotton I've added cluster token list/delete. PTAL.
> 
> > I just accidentally put 'tags' at the top level of a sandbox json file instead of under 'spec' and it accepted the request without error and silently dropped the tags. Should we check for that kind of thing?
> 
> Do you mean like a client-side schema check to see if you're trying to set an unknown field? That seems reasonable, but we'd probably want to put some thought into it as a separate effort/feature.

ok, sgtm.

enisoc:
Yes. I'm planning to do that after the API server-side release.

scott-cotton (approved):
LGTM

PR #10, opened 2022-08-08
publish linux docker images

scott-cotton:
This PR adds linux docker images (arm/amd64 so they should run on Mac/arm under docker) to the build/release.

### DIFF
--- a/goreleaser.yaml
+++ b/goreleaser.yaml
@@ -24,19 +24,19 @@ checksum:
   name_template: checksums.txt
 snapshot:
   name_template: "{{ .Version }}-{{ .ShortCommit }}"
-    #brews:
-    #- name: signadot-cli
-    #  description: Command-line interface for Signadot
-    #  homepage: https://signadot.com
-    #  folder: Formula
-    #  commit_author:
-    #    name: foxish
-    #    email: info@signadot.com
-    #  tap:
-    #    owner: signadot
-    #    name: homebrew-tap
-    #  install: |-
-    #    bin.install "signadot"
+brews:
+- name: signadot-cli
+  description: Command-line interface for Signadot
+  homepage: https://signadot.com
+  folder: Formula
+  commit_author:
+    name: foxish
+    email: info@signadot.com
+  tap:
+    owner: signadot
+    name: homebrew-tap
+  install: |-
+    bin.install "signadot"
 dockers:
   -
     use: buildx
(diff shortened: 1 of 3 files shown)


goreleaser.yaml
```diff
@@ -37,3 +37,47 @@ brews:
...
+    ids: ['signadot']
+    image_templates:
+      - "signadot/signadot-cli{{ .Env.SIGNADOT_IMAGE_SUFFIX }}:v{{ .Version }}-arm64v8"
+      - "signadot/signadot-cli{{ .Env.SIGNADOT_IMAGE_SUFFIX }}:latest-arm64v8"
+    build_flag_templates:
+      - "--label=com.signadot.image.created={{.Date}}"
+      - "--label=com.signadot.image.title={{.ProjectName}}"
+      - "--label=com.signadot.image.revision={{.ShortCommit}}"
+      - "--label=com.signadot.image.version={{.Version}}"
+      - "--platform=linux/arm64/v8"
```
foxish, line 70: Wondering if we need to make this available. Can docker on arm64 not run the amd64 image? Also, it seems rare to have a CI pipeline that uses MacOS. If we don't strictly need it, I'd rather we didn't build it.

scott-cotton: If I'm any indication of others, I often work on a Mac and test locally prior to putting into real CI, so I think it is a barrier to CI use to not provide it, whether or not the actual CI system is arm based.  For linux/arm systems, it would run more smoothly with the correct architecture.  This also makes the docker images run on the same hosts as the binary.  Are you sure you don't want to build it?

foxish: Can we leverage the multi-arch builds as detailed in https://docs.docker.com/desktop/multi-arch/#build-and-run-multi-architecture-images to just build one image that can support multiple architectures? This looked like it was building separate images to me.

scott-cotton: It looks like goreleaser needs to tag separate images per platform to accommodate manifests, but both docker buildx and goreleaser will need to actually build the different images, it is only a difference of visibility of arch specific tags.

If you'd like docker hub to only contain the visible non-hash tags for the multi-arch manifests, I think we'd have to either
1) remove the arch specific tags using something like https://devopsheaven.com/docker/dockerhub/2018/04/09/delete-docker-image-tag-dockerhub.html
or
2) use docker buildx+build manually on the images.  In this case, we couldn't pass multiple --platform args because each platform has a corresponding platform specific binary, so it wouldn't work exactly as in the linked docker buildx docs.

I tried setting skip_push to true on the arch-specific docker images and goreleaser pushes them anyway.

foxish (approved):
Based on a discussion we had offline - this looks good. The `:latest` and `:vN` images will work, and we just need to clean up the arch specific images later (if we want to). All user documentation will not use the architecture specific versions and just use `:latest` and `:vN`.

PR #12, opened 2022-08-12
fix docker amd/alpine problem

scott-cotton:
see
https://stackoverflow.com/questions/66963068/docker-alpine-executable-binary-not-found-even-if-in-path

Without this, running on docker with platform amd64 gave a very mysterious "not found" error for the executable.


foxish (approved):
LGTM, we want portable binaries here and can pay the size cost I think

PR #13, opened 2022-08-16
make delete only need to expand vars in name

scott-cotton:


foxish (approved):
LGTM, but I still think we should just encourage people to use `delete <name>` over `delete -f --set <name>` in our docs.

PR #16, opened 2022-09-02
reduce-wait-timeout

scott-cotton:
Following a discussion about the time it takes to discover a k8s-config error that occurs late and how this is inconvenient for interactive use, this PR proposing reducing the default wait timeout to 3m from 5m.

Rational: most clusters most of the time will probably succeed with 3m, but there will be some failures now and then due to performance bottlenecks.  With 5m, almost all  failures to become ready within the allotted time will be not be due performance bottlenecks.  5m might be a good in-CI default, but 1m would probably be best for interactive use.  Since users are discovering by interactive use, this seems to be a good compromise.  They can up the limit if it starts causing problems in CI due to cluster performance bottlenecks.


foxish (approved):
LGTM. Noting here that we'll need to add this to release notes since the default is changing.

PR #18, opened 2022-09-09
fix staticcheck and govulncheck problems, include staticcheck.conf

scott-cotton:


scott-cotton:
last commit fixes govulncheck issues.

foxish (approved):

PR #25, opened 2022-11-15
update to go-sdk-v0.3.4 (labels)

scott-cotton:


go.mod
```diff
@@ -4,9 +4,9 @@ go 1.19
 
 require (
 	github.com/docker/go-units v0.4.0
-	github.com/go-openapi/runtime v0.24.1
+	github.com/go-openapi/runtime v0.24.2
 	github.com/go-openapi/strfmt v0.21.3
-	github.com/signadot/go-sdk v0.3.3
+	github.com/signadot/go-sdk v0.3.4
```
jmsktm, line 9: I had pushed the v0.3.4 tag without merging in the code generated with the latest swagger.yaml. I did that afterwards and updated the tag to point to the latest commit with generated code. But pkg.go.dev still has the previous version at the moment.

https://pkg.go.dev/github.com/signadot/go-sdk@v0.3.4/models#SandboxSpec

image

With this, the sandboxes created with the newly generated CLI isn't getting created with labels (though the spec has labels defined).

jmsktm (approved):
As discussed on Slack, you seem to be getting the latest version of v0.3.4, whereas I'm getting the previous version from before re-tagging. Even cleaning local cache isn't helping.

Since you've able to create/get/list sandboxes with labels defined in the spec after this change, I suppose it's fine to go ahead with it. I'll deal with the cache issue on mine later.

PR #26, opened 2022-11-16
update go.sum to correct dependency v0.3.4

foxish:
A prior version of v0.3.4 was unintentionally pushed. It has now been removed and we're updating instead to the correct version of the go-sdk dependency.


jmsktm (approved):

foxish:
Still testing this PR to see if it fixes the dependency.

jmsktm (approved):
Approved for merge if the labels are working as expected.

foxish:
As stated in https://github.com/golang/go/issues/33969#issuecomment-528089385, this is not possible and we must tag a new version here.

PR #31, opened 2023-01-10
Addition of jsonexact.Unmarshal() to disallow unknown fields from YAML definitions

daniel-de-vera:
Changes required to address issue [#2539](https://github.com/signadot/signadot/issues/2539)


scott-cotton:
@daniel-de-vera do you think we should merge this before next release?

scott-cotton:
> 
> In any case, I think the future path of the CLI (based on Anirudh's view) is not going to rely on this sort of checking client-side, so things will change.
> 
> If it were up to me, I would include it in the release.

I'm not sure it's incompatible with moving toward a thin client.  

I think it can be included and would help.

cc @foxish wdyt?

scott-cotton (approved):
LGTM, but also asked for @foxish 's take

daniel-de-vera:
> > In any case, I think the future path of the CLI (based on Anirudh's view) is not going to rely on this sort of checking client-side, so things will change.
> > If it were up to me, I would include it in the release.
> 
> I'm not sure it's incompatible with moving toward a thin client.

No no, I don't think it is. What I was trying to say, is that writing a more complex logic than this client-side, to me is pointless given of the idea of moving toward a thin client.

scott-cotton:
> > > In any case, I think the future path of the CLI (based on Anirudh's view) is not going to rely on this sort of checking client-side, so things will change.
> > > If it were up to me, I would include it in the release.
> > 
> > 
> > I'm not sure it's incompatible with moving toward a thin client.
> 
> No no, I don't think it is. What I was trying to say, is that writing a more complex logic than this client-side, to me is pointless given the idea of moving toward a thin client.

Agreed, but also I don't think such logic would be specific to the client, rather it would be independent of where it is used.

PR #32, opened 2023-03-10
CRUD for Resource Plugins (v2)

jmsktm:
- **Includes:** CRUD for RP v2 (done)
- **TBD:** https://github.com/signadot/signadot/issues/2986


go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/docker/go-units v0.4.0
 	github.com/go-openapi/runtime v0.25.0
 	github.com/go-openapi/strfmt v0.21.3
+	github.com/signadot/go-sdk v0.3.7-0.20230310214135-3f47ce384c70
```
jmsktm, line 9: Result of `go get github.com/signadot/go-sdk@main` that points to the latest main branch which at the time of executing the command is https://github.com/signadot/cli/commit/eea80de6781fbcbdc06f64aab5d7e9484c1a787.

I'm not sure what the version here maps to `v0.3.7-0.20230310214135-3f47ce384c70`.  But I have verified that it works. If you know, pls lmk as well :)

jmsktm: My bad. That was the latest commit in the cli repo. The correct commit is https://github.com/signadot/go-sdk/commit/3f47ce384c70b41339c9a9665a6c5940e416277b. So the SHA is a match. 👍

internal/command/resourceplugin/subst.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+	for k := range topLevel {
+		if k == "name" {
+			continue
+		}
+		delete(topLevel, k)
+	}
+	return topLevel
+}
+
```
daniel-de-vera, line 46: @jmsktm, there is a lot of code duplication between all the functions below and the ones in `internal/command/sandbox/subst.go`.

I think it would be a good idea to locate this in some common package.

scott-cotton (commented):
I gave it a spin w/out templates and it seems ok so far.

+1 for trying to unify the substation a bit.

daniel-de-vera (approved):
Except for the comment I left before, it LGTM and think it works as expected (did some testing).

But because of https://github.com/signadot/signadot/issues/2984 (lack of the basePodTemplate) I couldn't really use the CLI yet to run realistic e2e tests.

scott-cotton:
one follow up about the templating ux

currently, with -o yaml, we sometimes get things that look like this:

```
      script: "#!/bin/bash\n# exit when any command fails\nset -e\n\necho \"Provisioning
        a MariaDB server\"\necho \"Sandbox id: ${SIGNADOT_SANDBOX_ID}\"\necho \"Resource
        name: ${SIGNADOT_RESOURCE_NAME}\"\necho \"DB name: ${DBNAME}\"\n\n# Install
        bitnami helm chart repo\nhelm repo add bitnami https://charts.bitnami.com/bitnami\n\n#
        Deploy a temporary DB for this Sandbox\nexport NAMESPACE=signadot\nRELEASE_NAME=\"signadot-${SIGNADOT_RESOURCE_NAME,,}-${SIGNADOT_SANDBOX_ID}\"\necho
        \"Installing Helm release: ${RELEASE_NAME}\"\nhelm -n ${NAMESPACE} install
        \"${RELEASE_NAME}\" bitnami/mariadb \\\n  --version 11.5.0 --wait --timeout
        5m0s \\\n  --set fullnameOverride=\"${RELEASE_NAME}-mariadb\" \\\n  --set
        primary.persistence.enabled=false \\\n  --set serviceAccount.create=false
        \\\n  --set auth.database=\"${DBNAME}\"\n  \n# Get the generated password,
        based on instructions from the Helm chart.\nMYSQL_ROOT_PASSWORD=$(kubectl
        -n ${NAMESPACE} get secret \"${RELEASE_NAME}-mariadb\" -o jsonpath=\"{.data.mariadb-root-password}\"
        | base64 -d)\n\n# Populate the outputs\nDBHOST=\"${RELEASE_NAME}-mariadb.${NAMESPACE}.svc\"\necho
        -n \"${DBHOST}\" > /tmp/host\necho -n 3306 > /tmp/port\necho -n \"${MYSQL_ROOT_PASSWORD}\"
        > /tmp/root-password\n"
```

and sometimes like this

```
 script: |
        #!/bin/bash
        # exit when any command fails
        set -e

        echo "Sandbox id: ${SIGNADOT_SANDBOX_ID}"
        echo "Resource name: ${SIGNADOT_RESOURCE_NAME}"
```

I'm not sure offhand how to control that but it looks worth thinking about to me:
- do we want an `-o template` option or something?
- is there a way to control whether a string is formatted with `|` or not with `-o yaml`?

jmsktm:
> I think it would be a good idea to locate this in some common package.

Done. Also updated the routegroups and sandboxes codes (respective subst.go) to use it.

> currently, with -o yaml, we sometimes get things that look like this

I too was getting that earlier. But I'm consistently getting a properly formatted script now. I'll merge this one in for now, and look into that separately.

PR #35, opened 2023-03-21
improved template support for variables and filepaths

scott-cotton:
This commit fixes 2 problems with the regexp based template parser:  variable names needed to match the same pattern as file paths, but the matching patterns are different, and the filepath matcher was too strict.

This commit permits filepaths which contain anything except '}' supposing that leading and trailing spaces are not part of the path.  I think a proper lexer would be needed to allow '}' (multiple expansions in one string trigger a problem for greedy `@{.*}`).

In addition, this commit extends the `[yaml]` argument into a generalized entity which permits raw, yaml, or binary encoding applied to either variables or operations like embed.  The idea is that the value taken from a variable or operation is subsequently _encoded_ into the document either as-is, as-yaml (including strings, numbers, arrays, dicts,...), or as-binary (base64 standard encoding as string, so as to map to Go json encoding of []byte).  By default, no `[encoding]` defaults to `[raw]`.  I hope to be able to extend the encoding to include
- including templates with expansion
- including directories

The parsing has been restructured to go over the input yaml and then replace strings.  Each string then is interpreted with  `@{...}` being expanded according to variables and files and encoding provided.  When the string being interpreted is exactly `@{...}`, encodings other than `raw` are permitted.


daniel-de-vera (approved):
LGTM

jmsktm (approved):
I see it's merged now, but I didn't have much in review other than a few nits related to readability.

As for the approach, I feel that while a single regex gets too complex as we extend our support for more templating features, working with regex in layers could take us all the way as much as character operations would in a cleaner and less error prone way. But I'm fine with this implementation as well.

PR #36, opened 2023-03-21
Use go-sdk v0.3.7

jmsktm:


foxish:
The field name changes don't result in some additional changes here to the models in use etc?

jmsktm:
> The field name changes don't result in some additional changes to the models in use etc? Just the dependency bump will suffice?

Yes. The content from the file will just map to the corresponding types referenced by the new version.

foxish (changes requested):
LGTM

foxish (approved):

PR #38, opened 2023-03-22
fix yaml output for multiline

scott-cotton:
always uses a yaml multiline syntax when the string contains multiple lines.

Fixes https://github.com/signadot/signadot/issues/2996


daniel-de-vera (approved):
LGTM

PR #41, opened 2023-06-22
Definition of the LocalNet & SandboxManager API specs

daniel-de-vera:
This PR proposes a spec for the Local Controller API and the Root Controller API.


scott-cotton (approved):
LGTM

PR #42, opened 2023-06-26
local command scaffolding and config

scott-cotton:
stacked on #41 

To be updated after discussion from todays meeting with new names for things in CLI scaffolding


internal/command/localdaemon/locald.go
daniel-de-vera, line 1: I would rename `localdaemon` to `locald` (that way it matches with the subcommand and also the package under internal).

internal/command/command.go
```diff
@@ -34,6 +35,7 @@ func New() *cobra.Command {
 		sandbox.New(cfg),
 		routegroup.New(cfg),
 		resourceplugin.New(cfg),
+		local.New(cfg),
```
daniel-de-vera, line 38: Why not making `localdaemon.newLocalDaemon()` public and add it in here?

daniel-de-vera (approved):
LGTM

scott-cotton:
note: will make the change to use the flag --sandbox-manager in another PR

PR #48, opened 2023-06-29
Upgrade of libconnect and fixes in local connect 

daniel-de-vera:


internal/command/local/connect.go
```diff
@@ -109,25 +109,27 @@ func runConnect(cmd *cobra.Command, cfg *config.LocalConnect, args []string) err
...
+					"locald",
+				)
 			} else {
 				cmdToRun = exec.Command(os.Args[0], "locald")
 				cmdToRun.Env = append(cmdToRun.Env,
-					fmt.Sprintf("SIGNADOT_LOCAL_CONNECT_INVOCATION_CONFIG=%s", string(ciBytes)),
 					fmt.Sprintf("HOME=%s", ciConfig.UIDHome),
 					fmt.Sprintf("PATH=%s", ciConfig.UIDPath))
 			}
+			cmdToRun.SysProcAttr = &syscall.SysProcAttr{
```
scott-cotton, line 125: I would have thought this should only apply to the sudo case.  It works fine for me without setsid in the unprivileged case.  wdyt?

scott-cotton (approved):
LGTM

PR #49, opened 2023-06-29
Functional version of Root Manager

daniel-de-vera:


scott-cotton:
I have tested this on Mac, I found this problem on control-c of `signadot local connect`:
```
Error: failed to terminate process: operation not permitted
```

I think on Macs we can't call process.Stop() for sudo to root.

The second thing is that we don't yet have this code https://github.com/signadot/signadot/blob/076486d2d172db9fda3bf16ff0f876bb1865432c/connect/pkg/sdctl/connect/outbound.go#L492 in place for Macs (it's the route part, it only does the pf part now)

daniel-de-vera:
> I have tested this on Mac, I found this problem on control-c of `signadot local connect`:
> 
> ```
> Error: failed to terminate process: operation not permitted
> ```
>
> I think on Macs we can't call process.Stop() for sudo to root.

Ok... I can implement the `shutdown()` method via the API and invoke it from `local connect`.
 
> The second thing is that we don't yet have this code https://github.com/signadot/signadot/blob/076486d2d172db9fda3bf16ff0f876bb1865432c/connect/pkg/sdctl/connect/outbound.go#L492 in place for Macs (it's the route part, it only does the pf part now)

Ok, I see. I wasn't aware of this.

PR #50, opened 2023-06-30
basic local locking

scott-cotton:
This provides locking of the local connect command

There is a hidden --clobber flag in case things go awry, as they inevitably do with persistent locks, but it normal operation it shouldn't be needed.


daniel-de-vera:
I was wondering if we can cross `libconnect/common/pmlock` with `libconnect/common/processes.IsPIDRunning()` avoiding the need for the clobber logic. Basically, store within the lock file the PID of the locker, and when locking always make sure that PID is running, otherwise automatically release the lock.

daniel-de-vera:
Another idea I had, was in `signadot local connect` bind to one specific port, and if we get `EADDRINUSE` it means there is a connect already running. Maybe it's not as robust as your approach, but could be enough?

scott-cotton:
> Another idea I had, was in `signadot local connect` bind to one specific port, and if we get `EADDRINUSE` it means there is a connect already running. Maybe it's not as robust as your approach, but could be enough?

I think that what we are locking is not really a good use case for using a port listen, because
/etc/hosts and the networking layering stuff is not related to a port.  If we use an existing port, then it could always be changed.    If we use a symbolic port, rather than one that is actually in use, then that feels like a misuse of resources to me.  If it's not an ephemeral port, then that is crowded and how do we guarantee it is free?  If it is an ephemeral port, then how do we guarantee it is free?

scott-cotton:
> I was wondering if we can cross `libconnect/common/pmlock` with `libconnect/common/processes.IsPIDRunning()` avoiding the need for the clobber logic. Basically, store within the lock file the PID of the locker, and when locking always make sure that PID is running, otherwise automatically release the lock.

I don't quite understand how this would work.  Can you elaborate?

scott-cotton:
closing, as locking was implemented differently in #52 

- as suggested above, don't consider it possible that the lock is held if the pid in the lock file is not running
- refactoring to support having the long-running processes lock and propagate any errors related to locking up the call chain
- ...

PR #51, opened 2023-07-03
Improvements and fixes on top of local-sandboxes-main

daniel-de-vera:


internal/locald/rootmanager/root_manager.go
```diff
@@ -132,9 +132,9 @@ func (m *rootManager) runSandboxManager(ctx context.Context) (err error) {
 			cmdToRun.Stdout = os.Stdout
 			cmdToRun.Stdin = os.Stdin
 			// Prevent signaling the children
-			//cmdToRun.SysProcAttr = &syscall.SysProcAttr{
-			//	Setpgid: true,
-			//}
+			cmdToRun.SysProcAttr = &syscall.SysProcAttr{
+				Setpgid: true,
+			}
```
daniel-de-vera, line 137: Without this, after `ctrl-c` the sandbox-manager was still running:

```bash
ddv@ddvpc:~/git/signadot/cli$ ps ax | grep "signadot local"
 212098 pts/0    Sl+    0:00 ./signadot local connect
 212110 ?        Ss     0:00 sudo -S --preserve-env=SIGNADOT_LOCAL_CONNECT_INVOCATION_CONFIG ./signadot locald
 212127 ?        Sl     0:00 ./signadot locald
 212140 ?        S      0:00 sudo -n -u #1000 --preserve-env=SIGNADOT_LOCAL_CONNECT_INVOCATION_CONFIG ./signadot locald
 212141 ?        Sl     0:00 ./signadot locald

ddv@ddvpc:~/git/signadot/cli$ ps ax | grep "signadot local"
 212141 ?        Sl     0:00 ./signadot locald
```

scott-cotton: I am unable to replicate this (on Mac). Also, it makes sense to me that the sandbox manager should be in the same process group as its caller, so that signals get sent to it when sent to its caller.  I guess on a Mac (or likely any BSD based system), when the parent exits it kills its children and on linux, not (at least under the Go libs we're using). 

However, since the `signadot local` invocation is the only thing getting a signal, I'd expect the root manager should shutdown its children regardless.  I'd suggest the root manager explicitly kill the sandbox manager on shutdown rather than relying on subtle platform dependent IPC.  I know we needed that for the root sudo but otherwise I think it's undesirable.

internal/locald/sandboxmanager/sandbox_monitor.go
```diff
@@ -35,28 +40,66 @@ func newSBMonitor(rk string, clapiClient clapiclient.Client, rtClient revtun.Cli
...
 
+func (sbm *sbMonitor) getStatus() *clapi.WatchSandboxStatus {
+	sbm.Lock()
+	defer sbm.Unlock()
+	return sbm.status
+}
+
+func (sbm *sbMonitor) stop() {
+	select {
+	case sbm.doneCh <- struct{}{}:
```
scott-cotton, line 60: I understand it is not needed in this particular case, but I'd still prefer to use `close()`, as it is more idiomatic and easier to evolve the code -- for example we could have to determine if it is closed after some code changes.

scott-cotton (approved):
Except for the 2 comments, LGTM.

internal/locald/rootmanager/root_manager.go
scott-cotton: following discussion, it seems like this is still about sudo (we had different versions) rather than MacOS.  Please consider resolved.

PR #55, opened 2023-07-07
Fix problem reading config in local connect, and some other improvements

daniel-de-vera:
Fix for, the following bug, reported by @jmsktm :

```
When I initially ran the below sandbox apply command:
signadot sandbox apply -f xxxx --config ~/.signadot/staging.yaml

- When there was a local component: It returned a 401 error
- When I removed the local component: the sandbox was applied successfully
```


jmsktm (approved):
LGTM

PR #56, opened 2023-07-10
support for macos vpn route commands

scott-cotton:
see https://github.com/signadot/libconnect/pull/25

Also includes some fixes/improvements:

- cluster is no longer part of ci-config (it is in the connection config)
- mechanism to pass/dump ci-config is improved


daniel-de-vera (approved):
LGTM

PR #57, opened 2023-07-10
minimal change to support new tolocal fields from cluster

scott-cotton:
this deliberately makes no attempt to consume the new info, as part of prioritising operator changes for release.

depends on https://github.com/signadot/signadot/pull/3438

and https://github.com/signadot/libconnect/pull/26


daniel-de-vera (approved):
LGTM

PR #61, opened 2023-07-12
Local status subcommand

daniel-de-vera:
Implementation of the `sigandot local status` command.
This is how it looks:

Screenshot from 2023-07-12 17-59-31


scott-cotton:
Another thing is that it looks like ciConfig.Unpriveleged is always true. Not sure what to do about that.

scott-cotton (approved):
LGTM

PR #62, opened 2023-07-13
local disconnect

scott-cotton:
stacked on #61 

- monitor until done
- handle unprivileged, and privileged modes
- spelling correction from libconnect
- clean pid files for disconnect for robustness when root manager is killed.


internal/command/local/disconnect.go
```diff
@@ -31,20 +36,93 @@ func runDisconnect(cfg *config.LocalDisconnect, args []string) error {
...
 }
+
+func (rs *runState) init(signadotDir string) bool {
+	rmPIDFile := filepath.Join(signadotDir, config.RootManagerPIDFile)
+	sbmPIDFile := filepath.Join(signadotDir, config.SandboxManagerPIDFile)
+	processes.CleanPIDFile(rmPIDFile)
+	processes.CleanPIDFile(sbmPIDFile)
+
+	rs.Root.C, rs.Root.PID, rs.Root.E = processes.MonitorPID(rmPIDFile, slog.Default())
+	rs.NotRoot.C, rs.NotRoot.PID, rs.NotRoot.E = processes.MonitorPID(sbmPIDFile, slog.Default())
```
daniel-de-vera, line 126: As far as I understand, during a disconnect process, we will be calling several times this function, and every time we call `processes.MonitorPID()` that's launching a new go runtime that will be running in background until the process ends. So, we would end up creating many goroutines here, am I right?

scott-cotton: yes I guess so.  thinking about a solution.

scott-cotton: updated

daniel-de-vera (approved):
LGTM

PR #65, opened 2023-07-14
Upgrade of libconnect - addressing localnet close issue

daniel-de-vera:
This is related with https://github.com/signadot/libconnect/pull/31

PR #66, opened 2023-07-14
Improvements to sandbox manager in locald

daniel-de-vera:
This PR includes:

1. https://github.com/signadot/cli/pull/65
2. When re-applying one sandbox it took about 10 seconds for the sanbox to be ready (this is related to closing the established tunnels during the update process).
3. Graceful termination of established tunnels during `local disconnect`.

### DIFF
--- a/internal/locald/sandboxmanager/grpc_server.go
+++ b/internal/locald/sandboxmanager/grpc_server.go
@@ -144,6 +144,10 @@ func (s *grpcServer) registerSandbox(sb *models.Sandbox) {
 	defer s.sbMu.Unlock()
 	sbm, present := s.sbMonitors[sb.Name]
 	if present {
+		// Close all current rev tunnels. When applying the sandbox in the SaaS,
+		// it will update the SDS in the target cluster, causing current tunnels
+		// to disconnect. By closing them before hand, we are more reactive.
+		sbm.closeRevTunnels()
 		// update the local spec
 		sbm.updateLocalsSpec(sb.Spec.Local)
 		return
@@ -261,4 +265,24 @@ func (s *grpcServer) sbStatuses() []*commonapi.SandboxStatus {
 		res = append(res, grpcStatus)
 	}
 	return res
+}
+
+func (s *grpcServer) stop() {
+	var wg sync.WaitGroup
+
+	// stop all sbx monitors
+	s.sbMu.Lock()
+	for _, sbMon := range s.sbMonitors {
+		wg.Add(1)
+		// overwrite the delete function
+		sbMon.delFn = func() {
+			defer wg.Done()
+		}
+		// stop the sbx monitor
+		sbMon.stop()
+	}
+	s.sbMu.Unlock()
+
+	// wait until all sbx monitor has stopped
+	wg.Wait()
 }
--- a/internal/locald/sandboxmanager/sandbox_manager.go
+++ b/internal/locald/sandboxmanager/sandbox_manager.go
@@ -31,6 +31,7 @@ type sandboxManager struct {
 	apiPort      uint16
 	hostname     string
 	grpcServer   *grpc.Server
+	sbmServer    *grpcServer
 	portForward  *portforward.PortForward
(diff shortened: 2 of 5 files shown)


internal/locald/sandboxmanager/grpc_server.go
```diff
@@ -144,6 +144,10 @@ func (s *grpcServer) registerSandbox(sb *models.Sandbox) {
 	defer s.sbMu.Unlock()
 	sbm, present := s.sbMonitors[sb.Name]
 	if present {
+		// Close all current rev tunnels. When applying the sandbox in the SaaS,
+		// it will update the SDS in the target cluster, causing current tunnels
+		// to disconnect. By closing them before hand, we are more reactive.
+		sbm.closeRevTunnels()
```
scott-cotton, line 150: It is a good observation that an SDS update triggers tunnel closure, and this would help.


But, I think that in light of this TODO https://github.com/signadot/cli/blob/5c8b8cbf11920940a66d468aeab93237e48bde20/internal/locald/sandboxmanager/sandbox_monitor.go#L190 (making updates work), we should not make that harder by closing tunnels here, rather, we should make the updates seamless in case of no changes and close the existing tunnel when there is a change.  I guess this was out of scope of your intended fixes in this PR. 

I have prepared and tested operator and CLI changes which takes the course of action mentioned above and in the TODO, will post shortly.

internal/locald/sandboxmanager/grpc_server.go
```diff
@@ -262,3 +266,23 @@ func (s *grpcServer) sbStatuses() []*commonapi.SandboxStatus {
...
+	// stop all sbx monitors
+	s.sbMu.Lock()
+	for _, sbMon := range s.sbMonitors {
+		wg.Add(1)
+		// overwrite the delete function
+		sbMon.delFn = func() {
+			defer wg.Done()
+		}
+		// stop the sbx monitor
+		sbMon.stop()
```
scott-cotton, line 282: I find this flow a bit hard to follow and would suggest
1. Adding the idea of a `doneAck` to `sbMon`
2. having `sbMon.stop` wait for the `<-doneAck`
3. keeping the `delFn` as is

daniel-de-vera: That would cause a deadlock because the `sbMon.delFn` locks `s.sbMu` too:

https://github.com/signadot/cli/blob/904e7b059b749c2b422f1d181dd2911c159e6c3a/internal/locald/sandboxmanager/grpc_server.go#L158-L160

The other idea I thought of was waiting until `len(s.sbMonitors) == 0`, but I liked more what I ended up implementing.
But I'm open to changing it if you think there's a better approach.

scott-cotton: I guess it's just inherently messy any which way.   So, your preference.

internal/locald/sandboxmanager/sandbox_monitor.go
```diff
@@ -210,6 +209,20 @@ func (sbm *sbMonitor) updateLocalsSpec(locals []*models.Local) {
 	sbm.triggerReconcile()
 }
 
+func (sbm *sbMonitor) closeRevTunnel(xwName string) {
+	revtun := sbm.revtuns[xwName]
+	if revtun == nil {
```
scott-cotton, line 214: when can this happen?

internal/locald/sandboxmanager/sandbox_manager.go
```diff
@@ -31,6 +31,7 @@ type sandboxManager struct {
 	apiPort      uint16
 	hostname     string
 	grpcServer   *grpc.Server
+	sbmServer    *grpcServer
```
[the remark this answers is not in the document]
daniel-de-vera: Yes, I like it

scott-cotton (approved):
LGTM, one question (perhaps for comments) and one rename suggestion (leftover, not introduced by this PR.)

PR #69, opened 2023-07-17
Addition of the --root-manager and  --sandbox-manager flags to locald

daniel-de-vera:
As part of this change, I took away `ConnectInvocationConfig.Unprivileged` which was something specific to the `locald` instance rather than to the whole run.

After connected you will find the following processes:

```bash
$ ps ax | grep "signadot local"
1270041 ?        Sl     0:00 ./signadot locald --root-manager
1270096 ?        Sl     0:00 ./signadot locald --sandbox-manager
```


scott-cotton (approved):
LGTM

PR #75, opened 2023-07-18
check sb cluster matches connected cluster

### LINKED ISSUE #73: Disallow creating local sandbox against cluster currently not connected
Occurs in the case where there are multiple k8s clusters - just leads to the sandbox hanging forever since the sdxw never shows up.

scott-cotton:
Here's the current error output
```
23-07-18 scott@air cli % ./signadot sandbox apply -f local-sb.yaml --set cluster=air
Error: unable to apply sandbox in sandboxmanager: rpc error: code = Internal desc = sandbox spec cluster "air" does not match connected cluster ("signadot-staging")
```

Closes #73


foxish (approved):
LGTM, thanks

PR #76, opened 2023-07-18
Remove strong kubectx dependency & improve status

daniel-de-vera:
These changes include the removal of the strong kubectx dependency (in the case of LB) and also improve `signadot local status` output (as per the discussion in https://signadot.slack.com/archives/C04ST9XTLTC/p1689624722091039.


scott-cotton (approved):
LGTM

foxish:
@jmsktm - with this change, the dashboard will have a usable "user" field being surfaced, so, we can display username@hostname on the dashboard.

PR #77, opened 2023-07-18
remove grpc layer errors from sandbox apply proxy

scott-cotton:
Removing grpc layer errors from sandbox apply proxy

Note my protoc version doesn't match and I'm not sure how to make it match, so someone with a matching protoc version will need to push appropriate changes.

Otherwise, this uses the `one of` error or content pattern, as grpc `status.Errorf` does not support wrapping and the only other alternative seemed quite heavyweight: using google api error .proto 

I don't think we need to map grpc errors here anymore...


daniel-de-vera:
I updated the protobuf files.

daniel-de-vera (approved):
I added some minor fixes in https://github.com/signadot/cli/pull/77/commits/69c7185c1ee9978534c1ca87308032bc40794744.
LGTM!

PR #80, opened 2023-07-19
Local baseline port int

scott-cotton:
incorporate local baseline port as int changes.

This uses go-sdk from branch local-baseline-port-int instead of creating a go-sdk PR because that seems weird to publish the go-sdk when it's not in swagger or the other sdks.


foxish (approved):
LGTM

PR #82, opened 2023-07-27
local connect --wait and --wait-timeout

scott-cotton:
when `--wait` is specified, connect will run until the connection is healthy or `--wait-timeout` is reached.  In the case `--wait-timeout` is reached before the connection is healthy, `signadot local connect --wait` will stop the background processes which are trying and failing to establish the connection and exit with a non-zero exit code.

also, disallow `-o` for `signadot local [dis]connect`, before this PR `-o yaml`, etc was a no-op


daniel-de-vera (approved):
LGTM

PR #86, opened 2023-09-14
fix ttl presentation for updatedAt

scott-cotton:
this also consumes a new go-sdk

Fixes https://github.com/signadot/community/issues/33

Note: this depends on go-sdk https://github.com/signadot/go-sdk/pull/33 and should not be merged until that has been merged and the go.mod updated.  However, we should be able to build it and test it independently with the go.mod in this PR as of now.


daniel-de-vera (approved):
LGTM

PR #87, opened 2023-10-11
add import for auth providers

scott-cotton:
from https://github.com/gruntwork-io/terratest/issues/976#issuecomment-1139051034

the imported package has no documentation, though.


daniel-de-vera (approved):
LGTM

foxish:
Before this is merged, we need an updated version of the libconnect dependency.

scott-cotton:
> Before this is merged, we need an updated version of the libconnect dependency.

done in last commit.  It required adding kubectl to top level required, as otherwise it was indirect and not downgrading.

PR #93, opened 2023-11-14
use libconnect fixes for /etc/hosts under docker

scott-cotton:
see https://github.com/signadot/libconnect/pull/39

and https://github.com/signadot/signadot/issues/3877


daniel-de-vera (approved):
LGTM

PR #92, opened 2023-11-09
Implementation of sandboxes watcher status

daniel-de-vera:
This is how it looks now:

```
$ signadot local connect
signadot local connect needs root privileges for:
	- updating /etc/hosts with cluster service names
	- configuring networking to direct local traffic to the cluster

signadot local connect has been started ✓
you can check its status with: signadot local status
```
```
$ signadot local status
* runtime config: cluster test, running with root-daemon
✗ Local connection not healthy!
* failed to run sandboxes watcher ("incompatible operator version, current CLI requires operator >= 0.15.0")
```

In a successful case:

```
$ signadot local status
* runtime config: cluster test, running with root-daemon
✓ Local connection healthy!
    * port-forward listening at ":46727"
    * localnet has been configured
    * 41 hosts accessible via /etc/hosts
    * sandboxes watcher is running                    <-------
* Connected Sandboxes:
    - daniel-hotrod-route
        * Routing Key: b7juzn2hsnl1j
        - local-route: routing from Deployment/route in namespace "hotrod"
            - remote port 8083 -> localhost:8083
        ✓ connection ready

```

PR #94, opened 2023-11-15
Add backwards compatibility (support old operators)

daniel-de-vera:
This PR adds support for old operators. It implements the following logic:

1. If `sandboxes-watcher` is running, the CLI will just apply the sandbox in the SaaS, `sandboxmanager` will discover the sandbox via the `WatchLocalSandboxes` stream.
2. If it is not running, the CLI will register the sandbox in `sandboxmanager` after it was created in the SaaS. In this case `sandboxmanager` will use the "old" `WatchSandbox` method from `tunnel-api` to monitor the sandbox.
3. We will always be trying to run `sandboxes-watcher` in background, so as soon as it starts running (e.g. someone upgrades the operator), it will switch from using `WatchSandbox` streams (one for each sandbox) to use `WatchLocalSandboxes`. This change will be transparent for all the connected tunnels.

Now, the watcher status is not used to define the health of the local connection, it will look like this, eg:

```
$ signadot local status
* runtime config: cluster test, running with root-daemon
✓ Local connection healthy!
    * port-forward listening at ":42003"
    * localnet has been configured
    * 34 hosts accessible via /etc/hosts
    * sandboxes watcher is not running ("this feature requires operator >= 0.14.1")
* Connected Sandboxes:
    - daniel-hotrod-route
        * Routing Key: 9cyzp485xj7q6
        - local-route: routing from Deployment/route in namespace "hotrod"
            - remote port 8083 -> localhost:8083
        ✓ connection ready
```

```
$ signadot local status
* runtime config: cluster test, running with root-daemon
✓ Local connection healthy!
    * port-forward listening at ":40905"
    * localnet has been configured
    * 36 hosts accessible via /etc/hosts
    * sandboxes watcher is running
* Connected Sandboxes:
    - daniel-hotrod-route
        * Routing Key: 9cyzp485xj7q6
        - local-route: routing from Deployment/route in namespace "hotrod"
            - remote port 8083 -> localhost:8083
        ✓ connection ready
```


internal/command/sandbox/apply.go
```diff
@@ -88,6 +89,17 @@ func apply(cfg *config.SandboxApply, out, log io.Writer, args []string) error {
 	fmt.Fprintf(log, "Created sandbox %q (routing key: %s) in cluster %q.\n\n",
 		req.Name, resp.RoutingKey, *req.Spec.Cluster)
 
+	if len(req.Spec.Local) > 0 && !sbmgr.IsWatcherRunning(status) {
+		// We don't really know here if it is a temporal problem or if we are
```
scott-cotton, line 93: do we want to add an operator version method in the tunnel api to help disambiguate cases like this?  if so, maybe add operator version to the status output?

daniel-de-vera: I can add it if we want, but I don't think it is strictly needed for this specific case.
We would have to check the version periodically in `locald`, because it may change over time.

scott-cotton: I think we will eventually need something like that in order to facilitate implementing compatibility in the client.  I understand it's not trivial b/c of the retries.  Not sure if you think it's worthwhile in the context of this PR, but I'd encourage it.  So, I'm ok if it's too much for this specific case, but maybe tracking adding that in an issue would help.

daniel-de-vera: I will implement this and https://github.com/signadot/cli/pull/92#issuecomment-1807949979, in a stacked PR on top of this one.

daniel-de-vera: @scott-cotton: here is the PR: https://github.com/signadot/cli/pull/95

scott-cotton (approved):
LGTM

PR #96, opened 2023-11-22
Clean local sandboxes on disconnect

daniel-de-vera:
Implementation of `signadot local disconnect --clean-local-sandboxes`

### DIFF
--- a/internal/command/local/disconnect.go
+++ b/internal/command/local/disconnect.go
@@ -2,6 +2,7 @@ package local
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


scott-cotton (commented):
This looks basically functional to me, but I wonder about whether it is worth it to do more to ensure it works in the face of possible concurrent sandbox applies.  Would it be safer to 1) disconnect and then 2) delete sandboxes with local machine id matching? 

(This is not a rhetorical question, I understand there are tradeoffs, just wondering what it means to clean local b/c locald may not be aware of all local sandboxes).  Maybe deleting sandboxes with local id matching could be done first as well.  

Another tricky thing to me about this is that by adding a flag to disconnect we seem to be implying that we implement the flag in the context of the connection persistent lock, but we don't.  Maybe a separate command would be more clear?

scott-cotton:
> We could try to perform a more "atomic" operation, or change the source of truth (get local sandboxes associated with our machine ID from SaaS API), but any of those options looks like overengined to me for this feature.

Could we then do the clean in a loop until GetStatus() returns no sandboxes?

scott-cotton (approved):
LGTM

we can work out corner cases of removal as you find appropriate.

daniel-de-vera:
> Could we then do the clean in a loop until GetStatus() returns no sandboxes?

@scott-cotton, I explored this option, but it's not as simple as it seems.
First, if I wait until all local sandboxes from status get removed, that means we have to wait for the full trip (CLI -> SaaS -> Operator -> CLI) and so we increase the likelihood that someone may create a new local sandbox in the meanwhile, and so it keeps running over and over.
Second, if we wait until no new local sandboxes are removed from the status (in status we find SB1, we remove SB1, check status and find again only SB1), it may very well happen that new sandboxes get created while we are removing, because the CLI will only find those local sandboxes after the full trip, so it's pretty much the same as no loop.

For now, I will leave it as it is.

PR #98, opened 2023-12-01
GHA for releaser

foxish:
Add GHA for releasing CLI

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,56 @@
+name: Release Signadot CLI
+on:
+  push:
+    tags:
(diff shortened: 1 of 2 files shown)


foxish:
> I guess this looks fine to me, although I'm not sure how to test it. Given the above point, I wonder if it wouldn't be better to get rid of `goreleaser` first, wdyt?

This was tested already in the action that was run here https://github.com/signadot/cli/actions/runs/7055151904/job/19205130658?pr=98 to push v0.6.0. IMO goreleaser still brings value and is worth having - because it's taking care of multiple artifacts and brew, and is extensible to include other types of packaging as well.

jmsktm (approved):
LGTM

PR #99, opened 2023-12-10
filter out excluded cidrs from /etc/hosts

scott-cotton:
depends on https://github.com/signadot/libconnect/pull/41


daniel-de-vera (approved):
LGTM

PR #102, opened 2024-01-31
add sandbox ready counter to list routegroup

davixcky:
Fixes https://github.com/signadot/signadot/issues/3393


davixcky:
PR updated with the changes on [main repo](https://github.com/signadot/signadot/issues/3393) discussion

scott-cotton (commented):
When you say "small fork" does that mean forked from somewhere?  If so, what's the license?

davixcky:
> When you say "small fork" does that mean forked from somewhere? If so, what's the license?
It doesn't explicitly say https://github.com/justincampbell/timeago

davixcky:
> > > When you say "small fork" does that mean forked from somewhere? If so, what's the license?
> > 
> > 
> > It doesn't explicitly say https://github.com/justincampbell/timeago
> 
> In that case, can we just use it as an import? Then we aren't redistributing and hence open to all the questions around distributing license-less copies.


We changed a little bit. We remove the "about" and "over"

scott-cotton:
> > > > When you say "small fork" does that mean forked from somewhere? If so, what's the license?
> > > 
> > > 
> > > It doesn't explicitly say https://github.com/justincampbell/timeago
> > 
> > 
> > In that case, can we just use it as an import? Then we aren't redistributing and hence open to all the questions around distributing license-less copies.
> 
> We changed a little bit. We remove the "about" and "over"

How about using https://pkg.go.dev/github.com/xeonx/timeago ?

scott-cotton (approved):
LGTM

PR #103, opened 2024-02-14
first working local ip mappings

scott-cotton:
Sharing this now as it completes my first end to end test, but leaving as draft to clean up and test more.

consumes the various libconnect PRs and operator tunnel-api PRs
(top of stack libconnect is https://github.com/signadot/libconnect/pull/47 and operator https://github.com/signadot/signadot/pull/4157)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/nsf/jsondiff v0.0.0-20210926074059-1e845ec5d249
 	github.com/panta/machineid v1.0.2
 	github.com/signadot/go-sdk v0.3.8-0.20231120112931-5aa0810b12b7
-	github.com/signadot/libconnect v0.1.1-0.20240215161123-08abced763ca
+	github.com/signadot/libconnect v0.1.1-0.20240216085049-66d396416c52
 	github.com/spf13/cobra v1.6.0
 	github.com/spf13/viper v1.11.0
 	github.com/theckman/yacspin v0.13.12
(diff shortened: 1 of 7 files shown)


daniel-de-vera:
@scott-cotton, honestly I have no idea how this sometimes works (sometimes I get timeouts when loading hotrod frontend), but I think this is conceptually wrong:

https://github.com/signadot/libconnect/blob/894d1f5e9ccb2cd4b42688244bca88b6e1135ba6/internal/socks5/conn_dialer.go#L50-L71

Basically, we are resolving domains locally on the CLI instead of doing so server-side on tunnel-proxy.
I thought we we going to use the [Socks5 ATYP=DOMAINNAME](https://datatracker.ietf.org/doc/html/rfc1928#section-4) option.
Do you agree?

scott-cotton:
> @scott-cotton, honestly I have no idea how this sometimes works (sometimes I get timeouts when loading hotrod frontend), but I think this is conceptually wrong:
> 
> https://github.com/signadot/libconnect/blob/894d1f5e9ccb2cd4b42688244bca88b6e1135ba6/internal/socks5/conn_dialer.go#L50-L71
> 
> Basically, we are resolving domains locally on the CLI instead of doing so server-side on tunnel-proxy. I thought we we going to use the [Socks5 ATYP=DOMAINNAME](https://datatracker.ietf.org/doc/html/rfc1928#section-4) option. Do you agree?

Yes agreed.  I don't know how it worked either.  In fact, I took it working as confirmation that we weren't doing that and so it got overlooked.

daniel-de-vera (approved):
LGTM

Issue #71, opened 2023-07-17
`local status -o X` should always emit valid yaml | json


foxish:
```
./signadot local status -o json
Error: signadot is not connected
```

This is not valid JSON, which will make it hard to stitch this command into a workflow

foxish:
cc @daniel-de-vera @scott-cotton

foxish:
Same issue exists on `connect` and `disconnect` as well.

scott-cotton:
It sounds like we might not have the same expectations, so I'd suggest discussing this issue more.

Here is what I'd expect for facilitating stitching this into a workflow:

1. `-o` refers to output, NOT stderr. 
2. `local status` should NOT consider being disconnected an command invocation error, nor should it consider an error in communicating with locald a command invocation error, rather, these should be part of the status and `local status` should almost never exit non-zero.
3. Upon command invocation error, the error messages are placed on stderr, and the output is not necessarily coherent. 
4. `local connect` and `local disconnect` _should_ return errors/exit non-zero when already connected or respectively already disconnected to help  the wrapper confirm the expected state.

For reference, [here](https://github.com/spf13/cobra/issues/914) is a related issue to using the cobra framework.

foxish:
@davixcky in case you missed it^

PR #104, opened 2024-02-14
wrap local commands to support different formats

davixcky:


internal/print/error.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/signadot/cli/internal/config"
+	"io"
+)
+
+func printErrorJson(err error) error {
+
+	return nil
+}
+
+func Error(out io.Writer, err error, outputFormat config.OutputFormat) error {
```
scott-cotton, line 14: Looks like this will change the exit status based on the output format, which is odd.  

Can we keep the CLI exit status non zero on error?  Perhaps using [`errors.Join`](https://pkg.go.dev/errors#Join) when json or yaml or the default switch case, combining the argument error with the error result of the encoding would not change the exit code.

davixcky: Screenshot 2024-02-14 at 12 08 48 PM

WDYT on this approach? 
Screenshot 2024-02-14 at 12 11 48 PM

scott-cotton:
@davixcky I added a comment here https://github.com/signadot/cli/issues/71#issuecomment-1945534887 to the related issue for discussing more what we want.

PR #105, opened 2024-02-19
Local wait default

scott-cotton:
This PR makes `--wait` the default for `signadot local connect` and adds the flag `--no-wait`

### DIFF
--- a/internal/config/local.go
+++ b/internal/config/local.go
@@ -1,8 +1,10 @@
 package config
 
 import (
+	"flag"
 	"fmt"
 	"os"
+	"strconv"
 	"time"
 
 	"github.com/signadot/libconnect/config"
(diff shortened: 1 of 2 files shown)


internal/config/local.go
```diff
@@ -88,12 +90,44 @@ func (c *LocalConnect) AddFlags(cmd *cobra.Command) {
...
 	cmd.Flags().BoolVar(&c.Unprivileged, "unprivileged", false, "run without root privileges")
-	cmd.Flags().BoolVar(&c.Wait, "wait", false, "wait for connection healthy")
+	cmd.Flags().BoolVar(&c.NoWait, "no-wait", false, "don't wait for connection healthy")
+	cmd.Flags().AddGoFlag(&flag.Flag{
+		Name: "wait",
+		Value: &waitFlagValue{
+			negPointer: &c.NoWait,
+		},
+		DefValue: "true",
+	})
```
daniel-de-vera, line 100: The output produced by the `--help` doesn't look right (the `--wait` description is misaligned):

```bash
$ signadot local connect --help
connect local machine to cluster

Usage:
  signadot local connect [flags]

Flags:
      --cluster string          specify cluster connection config
  -h, --help                    help for connect
      --no-wait                 don't wait for connection healthy
      --unprivileged            run without root privileges
      --wait waitFlag[=true]    
      --wait-timeout duration   timeout to wait (default 10s)

Global Flags:
      --config string   config file (default is $HOME/.signadot/config.yaml)
      --debug           enable debug output
  -o, --output string   output format (json|yaml)
```

daniel-de-vera: From the point of view of an user, I'm not sure if it's clear what's the default behavior here (will it wait or nor?):

```
      --no-wait                 don't wait for connection healthy
      --wait waitFlag[=true]    wait for the connection to become healthy
```
 
Also, I don't fully understand the meaning of `waitFlag[=true]`. Maybe we can hide the `--wait` flag?

scott-cotton: Looks like cobra supports AddGoFlag but not nicely.  I made it hidden.

daniel-de-vera (approved):
LGTM

PR #106, opened 2024-02-21
support no service local workloads

scott-cotton:
This depends on https://github.com/signadot/libconnect/pull/53

To have the actual new support, this also depends on and https://github.com/signadot/signadot/pull/4178 and
https://github.com/signadot/signadot/pull/4176

However, this should be harmless and simply continue to fail for no service local workloads on old operators or before the apiserver check change has merged.


daniel-de-vera (approved):
LGTM

PR #108, opened 2024-02-26
Implementation of the ControlPlaneProxy connection type

daniel-de-vera:
This is the implementation of the ControlPlaneProxy connection type.
Here is an example of a config file:

```
org: <your-org>
api_key: <your-api-key>
api_url: http://localhost:8080
proxy_url: http://localhost:8087
local:
  debug: true
  connections:
  - cluster: <cluster-name>
    type: ControlPlaneProxy
```

Connecting:

```
$ signadot local connect
signadot local connect needs root privileges for:
	- updating /etc/hosts with cluster service names
	- configuring networking to direct local traffic to the cluster
[sudo] password for ddv: 

signadot local connect has been started ✓
* runtime config: cluster test, running with root-daemon
✓ Local connection healthy!
    * operator version 0.15.1-web.5
    * control-plane proxy listening at ":39441"
    * localnet has been configured
    * 42 hosts accessible via /etc/hosts
    * sandboxes watcher is running
* Connected Sandboxes:
    - No active sandbox
```

Note that the status now also reports about the control-plane proxy.

FYI: I didn't make `ControlPlaneProxy` the default connection type in this PR, because I think it deserves some discussion. I added an entry for this in the umbrella issue,

This depends on https://github.com/signadot/libconnect/pull/55 and https://github.com/signadot/signadot/pull/4196


scott-cotton (approved):
LGTM

PR #109, opened 2024-02-27
Switch from proxy connect to local proxy

daniel-de-vera:
Move the proxy functionality from:

```bash
signadot proxy connect [--sandbox SANDBOX|--routegroup ROUTEGROUP|--cluster CLUSTER] --map <target-protocol>://<target-addr>|<bind-addr> ...
```

to:

```bash
signadot local proxy [--sandbox SANDBOX|--routegroup ROUTEGROUP|--cluster CLUSTER] --map <target-protocol>://<target-addr>|<bind-addr> ...
```


scott-cotton:
@daniel-de-vera 

since this hangs with old operators, I think we should add an operator version check in `proxy.signadot.com` and return an error which is surfaced on `local connect` (immediately) and `signadot local proxy`.

wdyt?

scott-cotton:
> @daniel-de-vera
> 
> since this hangs with old operators, I think we should add an operator version check in `proxy.signadot.com` and return an error which is surfaced on `local connect` (immediately) and `signadot local proxy`.
> 
> wdyt?


(This is more a global thing version skew ux problem and not related to any particular PR per se)

daniel-de-vera:
> since this hangs with old operators, I think we should add an operator version check in `proxy.signadot.com` and return an error which is surfaced on `local connect` (immediately) and `signadot local proxy`.
> 
> wdyt?

IMO, the problem is somewhat more trickier than that, because it's only related to Istio (an old operator without istio will work fine).
I guess we could detect that (by looking from the apiserver at the agent pods), but it's a bigger effort.

scott-cotton:
> > since this hangs with old operators, I think we should add an operator version check in `proxy.signadot.com` and return an error which is surfaced on `local connect` (immediately) and `signadot local proxy`.
> > wdyt?
> 
> IMO, the problem is somewhat more trickier than that, because it's only related to Istio (an old operator without istio will work fine). I guess we could detect that (by looking from the apiserver at the agent pods), but it's a bigger effort.

Yeah, that's why I suggested just version gating it.

scott-cotton (approved):
LGTM

scott-cotton:
> > > since this hangs with old operators, I think we should add an operator version check in `proxy.signadot.com` and return an error which is surfaced on `local connect` (immediately) and `signadot local proxy`.
> > > wdyt?
> > 
> > 
> > IMO, the problem is somewhat more trickier than that, because it's only related to Istio (an old operator without istio will work fine). I guess we could detect that (by looking from the apiserver at the agent pods), but it's a bigger effort.
> 
> Yeah, that's why I suggested just version gating it.

Let's continue this in https://github.com/signadot/signadot/issues/4180

daniel-de-vera:
> > > since this hangs with old operators, I think we should add an operator version check in `proxy.signadot.com` and return an error which is surfaced on `local connect` (immediately) and `signadot local proxy`.
> > > wdyt?
> > 
> > 
> > IMO, the problem is somewhat more trickier than that, because it's only related to Istio (an old operator without istio will work fine). I guess we could detect that (by looking from the apiserver at the agent pods), but it's a bigger effort.
> 
> Yeah, that's why I suggested just version gating it.

I think disabling proxy all together for an old operator is an overkill. This is only relevant for running `local connect`, but a user could perfectly use the proxy for accessing services other than the socks5 server. And in any case, the UX will be similar run running `local connect`, it will wait for some time and fail because it can't connect tunnel-proxy.

PR #112, opened 2024-03-04
Check version skew in local connect

daniel-de-vera:
E.g.:

```bash
$ signadot local connect
Error: The connection type ControlPlaneProxy requires operator > v0.15.0
```

### DIFF
--- a/internal/command/local/connect.go
+++ b/internal/command/local/connect.go
@@ -291,17 +291,23 @@ func checkVersionSkew(localConfig *config.LocalConnect, ciConfig *config.Connect
 	if err != nil {
 		return fmt.Errorf("error parsing cluster operator version, %w", err)
 	}
+	if operatorVer.Prerelease() != "" {
+		// this is a pre-release version, let's treat it as a stable version
+		ov, err := operatorVer.SetPrerelease("")
+		if err != nil {
+			return fmt.Errorf("error removing pre-release from cluster operator version, %w", err)
+		}
+		operatorVer = &ov
+	}
 
 	if ciConfig.ConnectionConfig.Type == connectcfg.ControlPlaneProxyLinkType {
-		// ControlPlaneProxy requires operator > 0.15.0, in the constraint below
-		// we use "> 0.15.1-0" instead of ">= 0.16.0" to catch our pre-release
-		// versions
-		cppConstraint, err := semver.NewConstraint("> 0.15.1-0")
+		// ControlPlaneProxy requires operator > 0.15.0
+		cppConstraint, err := semver.NewConstraint("> 0.15.0")
 		if err != nil {
 			return err // this shouldn't happen
 		}
 		if !cppConstraint.Check(operatorVer) {
-			return fmt.Errorf("The connection type ControlPlaneProxy requires operator > v0.15.0")
+			return fmt.Errorf("The connection type ControlPlaneProxy requires operator >= v0.16.0")
 		}
 	}
 	return nil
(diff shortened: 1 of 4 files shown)


internal/command/local/connect.go
```diff
@@ -264,6 +273,40 @@ doneWaiting:
...
+	if ciConfig.ConnectionConfig.Type == connectcfg.ControlPlaneProxyLinkType {
+		// ControlPlaneProxy requires operator > 0.15.0, in the constraint below
+		// we use "> 0.15.1-0" instead of ">= 0.16.0" to catch our pre-release
+		// versions
+		cppConstraint, err := semver.NewConstraint("> 0.15.1-0")
+		if err != nil {
+			return err // this shouldn't happen
+		}
+		if !cppConstraint.Check(operatorVer) {
+			return fmt.Errorf("The connection type ControlPlaneProxy requires operator > v0.15.0")
```
scott-cotton, line 304: I would prefer the message have `>=` instead of `>`.   We can also use server.Parse(v)+ .{Major,Minor,Patch}() as a cleaner way of checking in way consistent with our perversions.

scott-cotton (approved):
LGTM

left a minor comment.

I suppose we are not going to do the same thing for tcp:// in `signadot local proxy`?  might be better to include that.

internal/command/local/connect.go
daniel-de-vera: > I would prefer the message have >= instead of >

Sure, I can change it to ">= v0.16.0"

> We can also use server.Parse(v)+ .{Major,Minor,Patch}() as a cleaner way of checking in way consistent with our perversions.

I didn't get this point, can you explain more?

daniel-de-vera:
> I suppose we are not going to do the same thing for tcp:// in `signadot local proxy`? might be better to include that.

But that has nothing to do with the fix in the operator (the fix was only related for service tunnel-proxy). In other words, if someone tries connecting a TCP service inside the cluster (and they have Istio), what will determine if it works or not, is:

- the protocol characteristics (is it client first or server first, how verbose it is, etc)
- or if the target K8S service has an `appProtocol` defined or the port is named following the Istio recommendations.

internal/command/local/connect.go
scott-cotton: Given a semver `v`, we can check if it has a [`Prelease`](https://pkg.go.dev/github.com/Masterminds/semver/v3#Version.Prerelease) and if so, let it pass, otherwise, otherwise default to the major, minor, and patch components.  

Or, since we know the version history, we can strip the prerelease and then check `>` (leaving the error message with >=).

Personally, I find these alternatives more explicit and readable. 

(autocorrection changed `semver` to `server` which might have caused some confusion too)

scott-cotton:
> > I suppose we are not going to do the same thing for tcp:// in `signadot local proxy`? might be better to include that.
> 
> But that has nothing to do with the fix in the operator (the fix was only related for service tunnel-proxy). In other words, if someone tries connecting a TCP service inside the cluster (and they have Istio), what will determine if it works or not, is:
> 
> * the protocol characteristics (is it client first or server first, how verbose it is, etc)
> * or if the target K8S service has an `appProtocol` defined or the port is named following the Istio recommendations.

Oh, right, thanks. Probably worth documenting somewhere.

internal/command/local/connect.go
daniel-de-vera: @scott-cotton, is it better now? https://github.com/signadot/cli/pull/112/commits/e316dfef93fa47eaa0d744276ee1d1a45302242e

PR #113, opened 2024-03-04
fixes to unknown field handling

scott-cotton:
Unknown field handling prevents overwriting unknown fields with their default value, as can be problematic for example with TTL.

- add it to resource plugin
- make it apply only to spec


internal/command/resourceplugin/subst.go
```diff
@@ -17,13 +19,27 @@ func loadResourcePlugin(file string, tplVals config.TemplateVals, forDelete bool
...
+	switch x := un.(type) {
+	case map[string]any:
+		name, ok = x["name"].(string)
+		spec = x["spec"]
+	default:
+	}
+	if !ok {
+		return nil, errors.New("missing name or spec fields")
+	}
+	d, err := json.Marshal(spec)
```
daniel-de-vera, line 36: Can we place this logic into a function? (and call it from `resourceplugin`, `routegroup` and `sandbox`)

scott-cotton: sure, done.

daniel-de-vera (approved):
LGTM

PR #114, opened 2024-03-04
support ttl in rg

davixcky:
- add timeago to sbx and ttl for routegroup
- update go-sdk


daniel-de-vera:
I see the following problems:

1. Formatting issue (in xxx minutes **ago**):

With sandbox TTLs we have something like:

```bash
$ signadot sb get sandbox-better-route
...
TTL:              Tue, 05 Mar 2024 10:46:31 -03 (7 minutes)
```

In RG:

```
$ signadot rg get hotrod-rg
...
TTL:              Tue, 05 Mar 2024 10:48:52 -03 (in 10 minutes ago)
```

I guess ideally in both cases it would say something like `(in <offset> <unit>)`


2. Think we should improve the way we display RG TTLs. For example, I created a RG with:

```yaml
  ttl:
    duration: 10m
    offsetFrom: noMatchedSandboxes
```

When there is no sandbox matching the RG, the output looks good:

```bash
$ signadot rg get hotrod-rg
Name:             hotrod-rg
Routing Key:      udml6q4mtx81u
Cluster:          test
Created:          Mon, 04 Mar 2024 14:08:39 -03 (21 hours ago)
TTL:              Tue, 05 Mar 2024 10:58:15 -03 (in 10 minutes ago)
```

But when I create a sandbox matching the RG, the output is misleading (equivalent to the no TTL case):

```bash
$ signadot rg get hotrod-rg
Name:             hotrod-rg
Routing Key:      udml6q4mtx81u
Cluster:          test
Created:          Mon, 04 Mar 2024 14:08:39 -03 (21 hours ago)
TTL:              - (forever)
```

daniel-de-vera:
This is unrelated to this PR, but I see we have the same function defined in three different packages:

```golang
func formatTimestamp(in string) string {
	t, err := time.Parse(time.RFC3339, in)
	if err != nil {
		return in
	}
	elapsed := units.HumanDuration(time.Since(t))
	local := t.Local().Format(time.RFC1123)

	return fmt.Sprintf("%s (%s ago)", local, elapsed)
}
```

Maybe we can find a place in `utils` and use from there?

PR #115, opened 2024-03-05
use @ instead of | in proxy mapping specs

scott-cotton:
(e2e PR forthcoming... then I'll update the docs PRs)


daniel-de-vera (approved):
LGTM

PR #117, opened 2024-03-06
Addition of an installation script

daniel-de-vera:
Install the CLI with:

```sh
curl -sSLf https://raw.githubusercontent.com/signadot/cli/main/scripts/install.sh | sh
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,16 @@ NOTE: Starting next release, and hence in the main branch, this repository has a
 on a private repository, and hence building or running from source will not work.
 Previous releases should continue to work.
 
+## Install
+
+To install the CLI, run:
+
+```sh
+curl -sSLf https://raw.githubusercontent.com/signadot/cli/main/scripts/install.sh | sh
+```
+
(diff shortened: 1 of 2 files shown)


scott-cotton:
Was wondering if we could also suggest using `go install` as an alternative, seems desirable because

1. it does not require curl and ssl
2. it is a more reliable toolchain method because it does not run a script via a link.

I know `--version` is funny with this, but adding appropriate `-ldflags` to `go install`  can work (see [this Go issue](https://github.com/golang/go/issues/18369)

daniel-de-vera:
> Was wondering if we could also suggest using `go install` as an alternative

But from what I understand `go install` builds the binary locally, isn't it? In which case, how can we deal with `libconnect`?

scott-cotton:
> > Was wondering if we could also suggest using `go install` as an alternative
> 
> But from what I understand `go install` builds the binary locally, isn't it? In which case, how can we deal with `libconnect`?

doh, I forgot.  Maybe we should open-source libconnect.

PR #119, opened 2024-05-24
Support ict (jobs and jobrunnergroup)

davixcky:
Support to full JobRunnerGroup and work for Jobs still going (pending to see logs, download artifacts, cancel job)


internal/command/jobs/printers.go
```diff
@@ -0,0 +1,188 @@
...
+	fmt.Fprintf(out, "\nArtifacts\n")
+
+	if len(artifactsList) == 0 {
+		fmt.Fprintln(out, "No artifacts")
+		return nil
+	}
+
+	t := sdtab.New[jobArtifactRow](out)
+	t.AddHeader()
+	for _, artifact := range artifactsList {
```
foxish, line 181: I think we need not print each individual URL here at all - we can just print the dashboard URL once under `job get` and leave just the names of the artifact items here. We will use the `signadot artifact` command to enable downloading these entities anyway, so, the individual URLs don't help.

internal/command/jobs/printers.go
```diff
@@ -0,0 +1,188 @@
...
+
+func printJobDetails(cfg *config.Job, out io.Writer, job *models.JobsJob) error {
+	tw := tabwriter.NewWriter(out, 0, 0, 3, ' ', 0)
+
+	startedAt, duration := getStartedAndDuration(job)
+
+	fmt.Fprintf(tw, "Job Name:\t%s\n", job.Spec.NamePrefix)
+	fmt.Fprintf(tw, "Generated Job Name:\t%s\n", job.Name)
+	fmt.Fprintf(tw, "Status:\t%s\n", job.Status.Phase)
+	fmt.Fprintln(tw)
```
foxish, line 78: Why this separator?

foxish (commented):
Made a few small cosmetic changes. LGTM mostly except for a couple of comments

internal/config/root.go
```diff
@@ -80,3 +81,25 @@ func (c *Root) SandboxDashboardURL(id string) *url.URL {
...
+}
+
+func (c *Root) ArtifactDownloadUrl(org, jobName string, attemptID int64, fileName string) *url.URL {
+	u := url.URL{
+		Scheme:   "https",
+		Path:     path.Join("api.staging.signadot.com/api/v2/orgs/", org, "artifacts", "jobs", jobName, "attempts", strconv.FormatInt(attemptID, 10), "objects/download"),
+		RawQuery: "path=" + fileName,
+	}
+	return &u
+	//https://api.staging.signadot.com/api/v2/orgs/signadot/artifacts/jobs/first-job/attempts/first-job-0/objects/download?path=/tmp/test.txt
```
foxish, line 104: This comment can be removed?

internal/command/jobs/printers.go
davixcky: Screenshot 2024-05-27 at 9 04 17 AM

I though we want it as the Whimsical design

internal/command/logs/command.go
```diff
@@ -0,0 +1,78 @@
...
+		Scheme:   "https",
+		Path:     path.Join(strings.TrimPrefix(cfg.APIURL, "https://"), "api/v2/orgs", cfg.Org, "/jobs", cfg.Job, "/attempts/0/logs/stream"),
+		RawQuery: "type=" + cfg.Stream,
+	}
+
+	events := make(chan *sse.Event)
+
+	client := sse.NewClient(u.String())
+	client.Headers = map[string]string{
+		"authorization": "Bearer eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6InZwdkxCbEF3Ml9oMXRIOU5KbDlxaCJ9.eyJpc3MiOiJodHRwczovL2F1dGguc3RhZ2luZy5zaWduYWRvdC5jb20vIiwic3ViIjoiZ29vZ2xlLW9hdXRoMnwxMDAzNTM5OTY4MzMwMDU1Njk3MDAiLCJhdWQiOlsiaHR0cHM6Ly93b3Jrc3BhY2VzLWFwaXNlcnZlciIsImh0dHBzOi8vc2lnbmFkb3QtZGV2LnVzLmF1dGgwLmNvbS91c2VyaW5mbyJdLCJpYXQiOjE3MTY5MDU2OTEsImV4cCI6MTcxNjk5MjA5MSwic2NvcGUiOiJvcGVuaWQgcHJvZmlsZSBlbWFpbCIsImF6cCI6IkhYNlQ3UDk2cnViQ0IxQmJ4b2t1bTlpdDk0RWx5MkpCIn0.w3QeBKPmMgOzssm6ehSlglq2EEPFgabukSqegul1IsxZ_kuaKrF8rLsoEfLVX0q6NmTFUenlssIJ48NF1ekUA3534OWGI8bq_eL4p5RrAIpi0xyxqn3dBdxS002e0esGmdHPEYV9am1qdbrpuhxEqIdb6BtT3ZmueK696ficxv97ri66f7C0a7J9uud7kAdcz_vOr0URKOAbzKmk4D9iB139SyOThkEexlQ_KK-myCVgf0ecGMWDEaioQYbVUiQRHXyWGU2BkeCJIUF-FtJIz0gGxOQAuN49Ot4CtO12z0F28jBW2kSn9AT4_Rulnq1ZrXQieixVmxUss-yNAK7gZg",
```
jmsktm, line 48: Must be removed without trace in commits for merge.

jmsktm (dismissed):
Still reviewing. But blocking PR merge with `request changes` to prevent merge of this PR with the hard coded bearer token.

jmsktm (commented):
Went over the PR a little superficially and left some comments.

davixcky:
@jmsktm comments addressed, the only remains is the one that expect feedback from @foxish

internal/command/jobrunnergroup/apply.go
```diff
@@ -0,0 +1,70 @@
...
+		return err
+	}
+	resp := result.Payload
+
+	fmt.Fprintf(log, "Created runner %q (%q)\n\n", req.Name, cfg.RunnerGroupDashboardUrl(req.Name))
+
+	return writeOutput(cfg, out, resp)
+}
+
+func writeOutput(cfg *config.JobRunnerGroupApply, out io.Writer, resp *models.RunnergroupsRunnerGroup) error {
```
jmsktm, line 59: Let's tag the structs with `// @Name <name>` so that `RunnergroupsRunnerGroup` and `JobsJob` simply become `RunnerGroup` and `Job`.

Sample usage: https://github.com/signadot/signadot/blob/master/sandboxes/api/control/signadotv2/sandboxes/sandbox.go#L30

internal/command/jobrunnergroup/printers.go
```diff
@@ -0,0 +1,61 @@
...
+
+	if err := tw.Flush(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func readiness(status *models.RunnergroupsStatus) string {
+	return fmt.Sprintf("%d/%d pods ready", status.Pods.Ready, status.Pods.Ready+status.Pods.NotReady)
```
jmsktm, line 60: Looks like we'll need to account for idle as well.
image

Alternately, we should surface more details - including the counts for total, ready, not ready and idle. Sorry I hadn't accounted for this in the UX as the statuses were not clear at that point.

internal/command/jobs/printers.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	return t.Flush()
+}
+
+func printJobDetails(cfg *config.Job, out io.Writer, job *models.JobsJob) error {
+	tw := tabwriter.NewWriter(out, 0, 0, 3, ' ', 0)
+
+	startedAt, duration := getStartedAndDuration(job)
+
+	fmt.Fprintf(tw, "Job Prefix:\t%s\n", job.Spec.NamePrefix)
```
jmsktm, line 75: Job prefix may be less important to surface, I'd think.

internal/command/jobs/printers.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	if err := tw.Flush(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func getStartedAndDuration(job *models.JobsJob) (createdAtStr string, durationStr string) {
```
jmsktm, line 94: Let's rename the function to reflect creation (used below) instead of start.

internal/command/jobs/printers.go
```diff
@@ -0,0 +1,213 @@
...
+func printJobDetails(cfg *config.Job, out io.Writer, job *models.JobsJob) error {
+	tw := tabwriter.NewWriter(out, 0, 0, 3, ' ', 0)
+
+	startedAt, duration := getStartedAndDuration(job)
+
+	fmt.Fprintf(tw, "Job Prefix:\t%s\n", job.Spec.NamePrefix)
+	fmt.Fprintf(tw, "Job Name:\t%s\n", job.Name)
+	fmt.Fprintf(tw, "Status:\t%s\n", job.Status.Phase)
+	fmt.Fprintf(tw, "Environment:\t%s\n", getJobEnvironment(job))
+	fmt.Fprintf(tw, "Started At:\t%s\n", startedAt)
```
jmsktm, line 79: Created At

internal/command/jobs/printers.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+func printJobDetails(cfg *config.Job, out io.Writer, job *models.JobsJob) error {
+	tw := tabwriter.NewWriter(out, 0, 0, 3, ' ', 0)
+
+	startedAt, duration := getStartedAndDuration(job)
+
+	fmt.Fprintf(tw, "Job Prefix:\t%s\n", job.Spec.NamePrefix)
+	fmt.Fprintf(tw, "Job Name:\t%s\n", job.Name)
+	fmt.Fprintf(tw, "Status:\t%s\n", job.Status.Phase)
```
jmsktm, line 77: Wonder whether to call it `phase` (which is what it is as per the API) or `status` (keeping it in-line with statuses for other entities). @foxish ? I think the former might be preferable to alleviate confusion.

jmsktm (commented):
Left a couple more comments.

internal/command/jobrunnergroup/printers.go
davixcky: I also had Idle but @foxish make some changes

internal/command/artifact/command.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/signadot/cli/internal/config"
+	"github.com/spf13/cobra"
+)
+
+func New(api *config.API) *cobra.Command {
+	cfg := &config.Artifact{API: api}
+
+	cmd := &cobra.Command{
+		Use:   "artifact",
+		Short: "Inspect and manipulate artifact",
```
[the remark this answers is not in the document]
foxish: "Inspect and download artifacts"

internal/command/jobrunnergroup/apply.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/signadot/go-sdk/models"
+	"github.com/spf13/cobra"
+)
+
+func newApply(jobrunnergroup *config.JobRunnerGroup) *cobra.Command {
+	cfg := &config.JobRunnerGroupApply{JobRunnerGroup: jobrunnergroup}
+
+	cmd := &cobra.Command{
+		Use:   "apply -f FILENAME [ --set var1=val1 --set var2=val2 ... ]",
+		Short: "Create or update a jobrunnergroup with variable expansion",
```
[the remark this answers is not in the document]
foxish: That should be made consistent with routegroups and resourceplugins - I would say let's keep it consistent for now and if we want to change it as you suggested, do it wholesale across everything at a future point.

internal/command/jobrunnergroup/printers.go
foxish: I'd ignore idle, and just have ready / not-ready because those two are collectively exhaustive. Idle / not is a separate field and it's ok to not have that info right now.

internal/command/jobs/printers.go
foxish: I'm not sure we want to replace this - it seems to me like both are useful. The time of creation for how we order things by default, and the time of starting once we're looking at the job running.

jmsktm: I think the status used to have job start time only when it's in running phase, but apparently not so now. Having both sgtm.

internal/command/jobs/printers.go
jmsktm: sgtm

internal/command/jobrunnergroup/printers.go
jmsktm: ack

internal/command/jobrunnergroup/apply.go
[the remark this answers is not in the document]
jmsktm: if we are using that format in RGs and RPs, being consistent with that sgtm.

foxish (commented):
Ran through some of the test cases and it looks good to start. Only pending comment I have is that we should have `job get` output also reflect which jrg it's intending to run on. That's currently not visible. Rest of the behavior LGTM - @jmsktm - please approve once the code passes muster since you've been reviewing it more closely.

internal/command/logs/command.go
```diff
@@ -0,0 +1,88 @@
...
+				}
+
+				fmt.Print(log.Message)
+			case "error":
+				return errors.New(string(event.Data))
+			case "signal":
+
+				switch string(event.Data) {
+				case "EOF":
+					fmt.Println("Connection closed.")
```
foxish, line 77: This need not be printed - I think it can just print a new line and end.

internal/command/jobs/printers.go
```diff
@@ -0,0 +1,250 @@
...
+	resp, err := cfg.Client.Artifacts.ListJobAttemptArtifacts(params, nil)
+	if err != nil {
+		return []*models.JobArtifact{}, nil
+	}
+
+	return resp.Payload, nil
+}
+
+type jobArtifactRow struct {
+	Name string `sdtab:"NAME"`
```
foxish, line 198: It is PATH, not name

internal/command/jobs/printers.go
```diff
@@ -0,0 +1,250 @@
...
+	tw := tabwriter.NewWriter(out, 0, 0, 3, ' ', 0)
+
+	createdAt, duration := getCreatedAtAndDuration(job)
+
+	fmt.Fprintf(tw, "Job Name:\t%s\n", job.Name)
+	fmt.Fprintf(tw, "Status:\t%s\n", job.Status.Phase)
+	fmt.Fprintf(tw, "Environment:\t%s\n", getJobEnvironment(job))
+	fmt.Fprintf(tw, "Created At:\t%s\n", getCreatedAt(job))
+	fmt.Fprintf(tw, "Started At:\t%s\n", createdAt)
+	fmt.Fprintf(tw, "Duration:\t%s\n", duration)
```
jmsktm, line 99: Duration doesn't seem to be available until the job is complete. So maybe hide it when it's not set? Same for the other properties (eg. startTime).

image

internal/command/jobrunnergroup/command.go
```diff
@@ -0,0 +1,26 @@
...
+		Short:   "Inspect and manipulate jobrunnergroup",
+		Aliases: []string{"jrg"},
+	}
+
+	// Subcommands
+	cmd.AddCommand(
+		newGet(cfg),
+		newList(cfg),
+		newApply(cfg),
+		newDelete(cfg),
```
jmsktm, line 22: @foxish : I realized while testing that deleting a JRG deletes all the corresponding jobs as well. I wonder if we should require the user to pass a `--force` flag, or respond to a prompt so that the user is informed about this destructive activity.

internal/config/root.go
```diff
@@ -80,3 +81,24 @@ func (c *Root) SandboxDashboardURL(id string) *url.URL {
...
+func (c *Root) JobDashboardUrl(name string) *url.URL {
+	u := *c.DashboardURL
+	u.Path = path.Join(u.Path, "testing", "jobs", name)
+	return &u
+}
+
+func (c *Root) ArtifactDownloadUrl(org, jobName string, attemptID int64, fileName string) *url.URL {
+	u := url.URL{
+		Scheme:   "https",
+		Path:     path.Join("api.staging.signadot.com/api/v2/orgs/", org, "artifacts", "jobs", jobName, "attempts", strconv.FormatInt(attemptID, 10), "objects/download"),
```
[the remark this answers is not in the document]
foxish: FYI @davixcky - It should use the apiserver from the loaded configuration of the CLI

internal/command/jobrunnergroup/command.go
foxish: Not MVP scope, because there may be an actual `--force` option we add to not have the JRG drain gracefully.

internal/config/job.go
```diff
@@ -0,0 +1,42 @@
...
+
+type JobList struct {
+	*Job
+
+	// Flags
+	ShowCompleted bool
+}
+
+func (c *JobList) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().BoolVarP(&c.ShowCompleted, "show-completed", "", false, "List also jobs with completed status")
```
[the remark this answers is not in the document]
foxish: ack - I like this, maybe it's just `--all` then, and the default is queued + running.

internal/command/jobrunnergroup/command.go
jmsktm: ack

internal/config/root.go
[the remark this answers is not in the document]
davixcky: Removing this function since is not being used

jmsktm (approved):
Reviewed and tested locally. LGTM.

PR #120, opened 2024-06-04
fix environment field

davixcky:
This fixes https://github.com/signadot/signadot/issues/4461

### DIFF
--- a/internal/command/jobs/printers.go
+++ b/internal/command/jobs/printers.go
@@ -64,15 +64,7 @@ func printJobTable(cfg *config.JobList, out io.Writer, jobs []*models.Job) error
 
 		createdAt, duration := getAttemptCreatedAtAndDuration(job)
 
-		environment := ""
-		routingContext := job.Spec.RoutingContext
-		switch {
-		case routingContext == nil:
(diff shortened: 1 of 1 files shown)


internal/command/jobs/printers.go
```diff
@@ -173,15 +165,15 @@ func getAttemptCreatedAtAndDuration(job *models.Job) (createdAtStr string, durat
...
+	switch {
+	case routingContext == nil:
+	case len(routingContext.Sandbox) > 0:
+		return fmt.Sprintf("sandbox=%s", routingContext.Sandbox)
+	case len(routingContext.Routegroup) > 0:
+		return fmt.Sprintf("routegroup=%s", routingContext.Routegroup)
 	}
 
-	return fmt.Sprintf("%s (ROUTEGROUP)", routingContext.Routegroup)
+	return ""
```
jmsktm, line 176: Shouldn't this return `BASELINE`?

davixcky: To match UI yes, but we had empty in cli, @foxish any thoughts

scott-cotton (approved):
LGTM

PR #123, opened 2024-06-05
ict 4483 - add job attach to wait for job execution

davixcky:
> Fixes https://github.com/signadot/signadot/issues/4483


jmsktm (commented):
Looks mostly good to me. But the it seemed connected / attached for a while after the job was completed before disconnecting with the below message:

```
./signadot --config ~/.signadot/staging.yaml job submit -f control-backend-e2e.yaml --attach
Job control-backend-e2e-ptc0dxq queued on Job Runner Group: control-e2e

Dashboard page: https://app.signadot.com/testing/jobs/control-backend-e2e-ptc0dxq

Waiting for job execution
Cloning signadot repo
Running backend e2e tests
newman

Latest-Operator-Version

→ latest-operator-version (member user)
  GET https://api.staging.signadot.com/api/v1/vars/latest-operator-version
  200 OK ★ 521ms time ★ 1.13kB↑ 306B↓ size ★ 9↑ 7↓ headers ★ 0 cookies
  ┌ ↓ application/json ★ text ★ json ★ utf8 ★ 94B
...
logs
...

&{%!s(*models.JobsCanceledState=<nil>) %!s(*models.JobsCompletedState=&{0 }) %!s(*models.JobsQueuedState=<nil>) %!s(*models.JobsRunningState=<nil>)}
```

jmsktm (dismissed):
I saw the connection persist beyond job completion with the logs command too.

```
./signadot --config ~/.signadot/staging.yaml logs --job control-backend-e2e-wmmx22i
```

Even in this case, the connection stayed intact for a while (maybe 20/30 seconds) after the job completed.

Since this seems to be external to the scope of this PR, I will go ahead and approve it. 
This could be directly or indirectly related to https://github.com/signadot/signadot/issues/4484 as well. But let's create a separate ticket for the cli so that it doesn't get forgotten.

internal/command/jobs/printers.go
```diff
@@ -110,6 +113,59 @@ func printJobDetails(cfg *config.Job, out io.Writer, job *models.Job) error {
...
+		case "completed":
+			os.Exit(int(j.Status.Attempts[0].State.Completed.ExitCode))
+
+			return nil
+		case "queued":
+			if looped {
+				fmt.Fprintf(out, "\033[1A\033[K")
+			}
+
+			fmt.Fprintf(out, "Queued on Job Runner Group %s\n", j.Spec.RunnerGroup)
```
daniel-de-vera, line 138: This code may cause a DoS! Note that while the job is in `queued` state the CLI will send requests to the `apiserver` continuously!
We should retry periodically.

internal/command/jobs/printers.go
```diff
@@ -110,6 +113,59 @@ func printJobDetails(cfg *config.Job, out io.Writer, job *models.Job) error {
...
+
+	for {
+		j, err := getJob(cfg.Job, jobName)
+		if err != nil {
+			return err
+		}
+
+		switch j.Status.Phase {
+		case "completed":
+			os.Exit(int(j.Status.Attempts[0].State.Completed.ExitCode))
```
daniel-de-vera, line 130: Not sure how we end up with `ExitCode int64` in the SDK, but both in the operator and the SaaS, the exit code of a given job may be nil:

https://github.com/signadot/signadot/blob/7740cc65653ed190cf089bd463df11e7aa9e7563/sandboxes/api/control/signadotv2/jobs/status.go#L56-L60

I would double check, I think there is the risk of treating `nil` exit codes as `0` (when a `nil` exit code, means there was an error running the job).

internal/command/jobs/printers.go
jmsktm: Right. Missed it during review.

daniel-de-vera:
I think the `--attach` feature should be more robust (at least if we want to use it in the context of CI). Something that in case of errors it retries based on exp backoff (or luby), and only after some time it fails. Otherwise we may end up with generating false negatives (claiming that a job failed when it doesn't).

For example, imagine the case when we restart the `apiserver` (e.g. we perform a new control release), all connected clients will error out, but that doesn't mean the Job failed.

internal/command/jobs/printers.go
davixcky: 10 seconds before each retry or increasing retry??

daniel-de-vera: Yes, 10s sgtm.

internal/command/jobs/printers.go
```diff
@@ -119,6 +125,79 @@ func printJobDetails(cfg *config.Job, out io.Writer, job *models.Job) error {
...
+				fmt.Fprintf(out, "\033[1A\033[K")
+			}
+
+			fmt.Fprintf(out, "Queued on Job Runner Group %s\n", j.Spec.RunnerGroup)
+		case "running":
+			if looped {
+				fmt.Fprintf(out, "\033[1A\033[K")
+			}
+
+			lastCursor, err = logs.ShowLogs(ctx, cfg.API, out, jobName, cfg.Attach, lastCursor, 0)
```
daniel-de-vera, line 172: If there multiple `logs.ShowLogs()` errors in a row, this code is resetting the cursor.
We should only update `lastCursor` is `err == nil`.
e.g.:

```
Welcome 27 times (env TEST=)
Welcome 28 times (env TEST=)
Welcome 29 times (env TEST=)
Welcome 30 times (env TEST=)
Welcome 31 times (env TEST=)
Welcome 32 times (env TEST=)
Welcome 33 times (env TEST=)
Welcome 34 times (env TEST=)
Welcome 35 times (enHello this is the demo meeting
Welcome 1 times (env TEST=))                                 Error getting logs: 502 Bad Gateway: no active tunnel found for cluster
Welcome 2 times (env TEST=)no active tunnel found for clusterp://localhost:8080/api/v2/orgs/ddvcorp/jobs/basic-aqzp1ua": dial tcp 127.0.0.1:8080: connect: connection refused
Welcome 3 times (env TEST=)
Welcome 4 times (env TEST=)
Welcome 5 times (env TEST=)
Welcome 6 times (env TEST=)
Welcome 7 times (env TEST=)
Welcome 8 times (env TEST=)
Welcome 9 times (env TEST=)
Welcome 10 times (env TEST=)
```

internal/command/jobs/printers.go
```diff
@@ -119,6 +125,79 @@ func printJobDetails(cfg *config.Job, out io.Writer, job *models.Job) error {
...
+
+			if err != nil {
+				fmt.Fprintf(out, "Error getting logs: %s", err.Error())
```
daniel-de-vera, line 175: We should add a newline char at the end of string.

internal/command/jobs/printers.go
```diff
@@ -119,6 +125,79 @@ func printJobDetails(cfg *config.Job, out io.Writer, job *models.Job) error {
...
+			os.Exit(int(attempt.State.Succeeded.ExitCode))
+
+			return true
+		case "queued":
+			if looped {
+				fmt.Fprintf(out, "\033[1A\033[K")
+			}
+
+			fmt.Fprintf(out, "Queued on Job Runner Group %s\n", j.Spec.RunnerGroup)
+		case "running":
```
daniel-de-vera, line 167: When a run completes we are always adding an extra 10 sec delay before terminating the execution (because the logs will terminate and we will go thru the retry process before finding out that the job has completed).

I would get the job once again after `logs.ShowLogs()` has completed (independent from the result), and if the job is in a terminal state end the execution there. Something like:

```golang
case "running":
	if looped {
		fmt.Fprintf(out, "\033[1A\033[K")
	}

	lastCursor, err = logs.ShowLogs(ctx, cfg.API, out, jobName, cfg.Attach, lastCursor, 0)
	if err != nil {
		fmt.Fprintf(out, "Error getting logs: %s", err.Error())
	}

	if j, err = getJob(cfg.Job, jobName); err == nil {
		switch j.Status.Attempts[0].Phase {
		case "failed":
			os.Exit(int(attempt.State.Failed.ExitCode))
			return true
		case "succeeded":
			return true
		}
	}
```

internal/command/jobs/printers.go
```diff
@@ -119,6 +125,79 @@ func printJobDetails(cfg *config.Job, out io.Writer, job *models.Job) error {
...
+			fmt.Fprintf(out, "Error getting job: %s", err.Error())
+
+			// We want to keep retrying if the timeout has not been exceeded
+			return false
+		}
+
+		attempt := j.Status.Attempts[0]
+		switch attempt.Phase {
+		case "failed":
+			os.Exit(int(attempt.State.Failed.ExitCode))
```
daniel-de-vera, line 153: We should update the go-sdk to the latest version.
In that one the `Failed.ExitCode` became a pointer.

internal/config/job.go
```diff
@@ -13,13 +14,36 @@ type JobSubmit struct {
...
+	if !attach.Changed {
+		return true
+	}
+
+	switch attach.Value.String() {
+	case "stdout", "stderr":
+		return true
+	}
+
+	return false
```
daniel-de-vera, line 46: I'm not sure this is working as expected, e.g.:

```
$ go run ./cmd/signadot/ job submit -f /home/ddv/git/signadot/signadot/sandboxes/test/control/data/jobs/basic.yaml --attach stdout
Error: unknown command "stdout" for "signadot job submit"
exit status 1

```

internal/config/job.go
```diff
@@ -13,13 +14,36 @@ type JobSubmit struct {
...
 }
 
 func (c *JobSubmit) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVarP(&c.Filename, "filename", "f", "", "YAML or JSON file containing the jobs creation request")
 	cmd.MarkFlagRequired("filename")
 	cmd.Flags().Var(&c.TemplateVals, "set", "--set var=val")
+
+	cmd.Flags().StringVarP(&c.Attach, "attach", "", "", "Waits until the job runs exits. Accept stdout or stderr as param")
+
+	cmd.Flags().DurationVar(&c.Timeout, "timeout", 0, "Timeout when waiting any retry or connection fails. If timeout is 0, will wait for completion or cancellation")
```
daniel-de-vera, line 29: I would rewrite like `timeout when waiting for the job to be started, if 0 is specified, no timeout will be applied and the command will wait until completion or cancellation of the job (default 0)`

internal/config/job.go
davixcky: $ go run ./cmd/signadot/ job submit --attach=stdout


That's the proper usage

daniel-de-vera: Oh, that's strange, why the other flags works without the "=" and not this one?.
E.g.:

```bash
$ go run ./cmd/signadot/ job get basic-0xgdwza --output yaml
createdAt: 2024-06-14T12:11:41Z
name: basic-0xgdwza
spec:
...
```

```bash
$ go run ./cmd/signadot/ job get basic-0xgdwza --output=yaml
createdAt: 2024-06-14T12:11:41Z
name: basic-0xgdwza
spec:
...
```

scott-cotton: I think it is because `--attach` has an optional argument but `--output` does not

davixcky: Correct, this is the reference to the docs
https://github.com/spf13/pflag?tab=readme-ov-file#command-line-flag-syntax

internal/command/jobs/printers.go
davixcky: How can i reproduce this? I tried turning off my wifi, but i kept the connection open with the sse stream

internal/command/jobs/printers.go
davixcky: Agree, also i removed the "failed" and "succeeded" from the all cases, since that case will never happen

internal/command/jobs/printers.go
daniel-de-vera: You can disconnect your cluster, and/or restart the `apiserver` (note that in my case, I'm using a local version).

daniel-de-vera (approved):
LGTM

PR #124, opened 2024-06-07
ict 4505

davixcky:
Fixes https://github.com/signadot/signadot/issues/4505


internal/command/artifact/utils.go
```diff
@@ -0,0 +1,40 @@
+package artifact
+
+import "github.com/go-openapi/runtime"
+
+func getDownloadConsumers() map[string]runtime.Consumer {
+	return map[string]runtime.Consumer{
+		"image/png":          runtime.ByteStreamConsumer(),
```
jmsktm, line 7: I had done a similar mapping elsewhere to correct the extension based on the MIME type (because in one case, an image was being uploaded with .txt). But I find it limiting to do it for just the considered extensions. So I think I'd want to consider skipping extension correction altogether to not try to correct for the user error.

Maybe we can do the same here?

daniel-de-vera: I agree with @jmsktm, we should try to skip the need to define this mapping, and basically map everything to `runtime.ByteStreamConsumer()`.

Here is what I think it could be a solution.
We could convert this code:

https://github.com/signadot/go-sdk/blob/15dad2127b066da9b9cae775a1a8ba2aac06716d/transport/transport.go#L27-L30

To:

```golang
type APIConfig struct {
	// Advanced (optional) settings
	Consumers          map[string]runtime.Consumer
	OverwriteConsumers bool
	Producers          map[string]runtime.Producer
	OverwriteProducers bool
	HTTPClient         *http.Client
}
```

Then, convert this code:

https://github.com/signadot/go-sdk/blob/15dad2127b066da9b9cae775a1a8ba2aac06716d/transport/transport.go#L89-L95

To:

```golang
	// apply consumer and producer settings
	if conf.OverwriteConsumers {
		rt.Consumers = conf.Consumers
	} else {
		for mimeType, consumer := range conf.Consumers {
			rt.Consumers[mimeType] = consumer
		}
	}
	if conf.OverwriteProducers {
		rt.Producers = conf.Producers
	} else {
		for mimeType, producer := range conf.Producers {
			rt.Producers[mimeType] = producer
		}
	}
```

And finally use `*/*` as the mime type in the consumer overwrite:

https://github.com/go-openapi/runtime/blob/9be04f2eaa9d1f69a45413ce15469d3ce8dcf990/client/runtime.go#L520-L527

daniel-de-vera: @davixcky, as discussed, I implemented this changes in https://github.com/signadot/cli/pull/127.

davixcky:
@daniel-de-vera @jmsktm i think we can close this one

jmsktm:
Yes, let's close this as Daniel's PR has already addressed it.

PR #129, opened 2024-06-13
use go-sdk w/out tries

scott-cotton:
depends on https://github.com/signadot/go-sdk/pull/50

Proposing this as an alternative

For reference, please see https://github.com/signadot/signadot/pull/4552

Fixes https://github.com/signadot/signadot/issues/4541


scott-cotton:
already incorporated

Issue #136, opened 2024-06-28
[Job] --attach hangs when job ends too quickly


foxish:
I have a job that's just:

```
spec:
  namePrefix: my-job
  runnerGroup: hackspace-test-jobs
  script: |
    #!/bin/bash

    echo abc
```

This just hangs on the CLI even though the job has already completed. 
It works as expected when I specify `sleep 10`.

cc @daniel-de-vera @davixcky

PR #138, opened 2024-06-28
fix typo submit

davixcky:
Fixes https://github.com/signadot/cli/issues/136

But i also notice one issue, since the job is really short, the logs are not print out


foxish (approved):
LGTM

PR #151, opened 2024-12-19
add check to access agent-metrics endpoint

scott-cotton:
This is intended as minimal change to fix

https://github.com/signadot/signadot/issues/5172

tested with warp, it causes restarts and we are able to connect whereas previously it hangs

regarding discussion around:

- do we need to make this visible in status?
I changed it so that it is more visible in status, returning unhealthy when a restart is needed instead of healthy.

- do we need to make check period configurable?

I think not, because it is actually complicated because there 2 check periods for healthy and unhealthy state.  Also, the check period relates to the timeout for the added checking request.

### DIFF
--- a/internal/locald/rootmanager/tp_monitor.go
+++ b/internal/locald/rootmanager/tp_monitor.go
@@ -2,6 +2,7 @@ package rootmanager
 
 import (
 	"fmt"
+	"net/http"
 	"time"
 
 	"log/slog"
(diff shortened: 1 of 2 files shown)


internal/locald/rootmanager/tp_monitor.go
```diff
@@ -128,6 +129,23 @@ func (mon *tpMonitor) checkTunnelProxyAccess(ctx context.Context) bool {
...
 		}
 	}
+	if !restartSvcs {
+		// the grpc check for connecting to the tunnel proxy does not suffice
+		// because it has built-in retries and may re-use a connection while
+		// we are unable to establish a new connection.  So, we also check
+		// the agent health endpoint
+		cli := &http.Client{
+			Transport: &http.Transport{},
+		}
```
daniel-de-vera, line 139: Maybe it is worth configuring a timeout here?

daniel-de-vera (approved):
LGTM, just left one comment.

internal/locald/rootmanager/tp_monitor.go
scott-cotton: I added a timeout and made a couple of other changes and fixes:
1. access agent-metrics (we don't have any health endpoints exposed, this was always restarting)
2. make the health more pessimistic, return unhealthy when restart is needed

scott-cotton:
> LGTM, just left one comment.

thanks, I made a couple of changes after more testing.  could you look again?

daniel-de-vera:
> > LGTM, just left one comment.
> 
> thanks, I made a couple of changes after more testing. could you look again?

Yes, still LGTM.

scott-cotton:
I have updated this to condition the additional check on the readiness of etc hosts and it now seems to be working fine.

internal/locald/rootmanager/tp_monitor.go
```diff
@@ -128,16 +129,40 @@ func (mon *tpMonitor) checkTunnelProxyAccess(ctx context.Context) bool {
 			restartSvcs = true
 		}
 	}
+	if !restartSvcs {
+		if mon.root == nil || mon.root.root == nil || mon.root.root.etcHostsSVC == nil || !mon.root.root.etcHostsSVC.Status().Healthy {
+			return false
+		}
+	}
+	if !restartSvcs {
```
daniel-de-vera, line 137: I think this block of code could be part of the same `if` statement from line 132.

internal/locald/rootmanager/tp_monitor.go
```diff
@@ -128,16 +129,40 @@ func (mon *tpMonitor) checkTunnelProxyAccess(ctx context.Context) bool {
 			restartSvcs = true
 		}
 	}
+	if !restartSvcs {
+		if mon.root == nil || mon.root.root == nil || mon.root.root.etcHostsSVC == nil || !mon.root.root.etcHostsSVC.Status().Healthy {
```
daniel-de-vera, line 133: 1. Why are we only checking for the health of `etcHostsSVC`, instead of doing it also for `localnetSVC`?
I think both should be healthy before trying to exercise an in-cluster request.

2. Nit: we are already assuming in the code from above that `mon.root != nil`, so I don't think we need that check.

internal/locald/rootmanager/tp_monitor.go
scott-cotton: sure

internal/locald/rootmanager/tp_monitor.go
scott-cotton: > Why are we only checking for the health of etcHostsSVC, instead of doing it also for localnetSVC?
I think both should be healthy before trying to exercise an in-cluster request.

I've added the check.  I think they should have been checked whether or not we exercise a fresh in-cluster request -- that is likely a cause of the status reporting healthy when it is not.

anyway, thanks for the review, I've revised it.

internal/locald/rootmanager/tp_monitor.go
```diff
@@ -128,16 +129,45 @@ func (mon *tpMonitor) checkTunnelProxyAccess(ctx context.Context) bool {
 			restartSvcs = true
 		}
 	}
+	if !restartSvcs {
+		rootMgr := mon.root.root
+		if rootMgr == nil {
+			return false
+		}
+		if rootMgr.localnetSVC == nil || rootMgr.localnetSVC.Status().Healthy {
```
daniel-de-vera, line 137: It should be `!rootMgr.localnetSVC.Status().Healthy`, isn't it?

scott-cotton: oops, thx!

daniel-de-vera:
LGTM, take 3! 😅

PR #157, opened 2025-03-18
repoconfig library to read git directories and .labels files

foxish:
This addresses https://github.com/signadot/signadot/issues/5546. Adds a package that helps read git config and labels from a repo. It is hooked up to `signadot test run` for now (when run with no args). Example run below.

```
go run ./cmd/signadot/main.go test run
Reading config from: /Users/anirudhramanathan/Documents/cli/.signadot/config.yaml
Found 4 test files:
  tests/fixtures/smart-tests-git/backend/team1/file1.star
    Labels:
      suite=integration
      area=backend
      team=team1
      signadot.com/run-id=skthedwn
  tests/fixtures/smart-tests-git/backend/team2/file2.star
    Labels:
      area=backend
      team=team2
      signadot.com/run-id=skthedwn
      suite=integration
  tests/fixtures/smart-tests-git/backend/test.star
    Labels:
      suite=integration
      area=backend
      signadot.com/run-id=skthedwn
  tests/fixtures/smart-tests-git/frontend/file3.star
    Labels:
      suite=integration
      team=frontend
      signadot.com/run-id=skthedwn
```

### DIFF
--- a/internal/repoconfig/config.go
+++ b/internal/repoconfig/config.go
@@ -173,9 +173,24 @@ func FindTestFiles(startPath string, cfg *Config) ([]TestFile, error) {
 	}
 	sortedPaths := sortPathsByLength(cleanPaths)
 
-	for _, testsDir := range sortedPaths {
-		if err := walkTestDirectory(repoRoot, testsDir, dirLabelsCache, runID, testFileMap); err != nil {
-			return nil, err
+	for _, path := range sortedPaths {
+		absPath := filepath.Join(repoRoot, path)
+		fileInfo, err := os.Stat(absPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to stat path %q: %w", path, err)
+		}
+
+		if fileInfo.IsDir() {
+			if err := walkTestDirectory(repoRoot, path, dirLabelsCache, runID, testFileMap); err != nil {
+				return nil, err
+			}
+		} else {
+			if !isTestFile(absPath) {
+				continue
+			}
+			if err := processTestFile(repoRoot, absPath, dirLabelsCache, runID, testFileMap); err != nil {
+				return nil, err
+			}
 		}
 	}
 
--- a/internal/repoconfig/config_test.go
+++ b/internal/repoconfig/config_test.go
@@ -124,3 +124,57 @@ func TestFileDiscovery(t *testing.T) {
 		}
 	}
 }
+
+func TestDirectFilePathInConfig(t *testing.T) {
+	// Create a config with only a direct file path
+	cfg := &Config{
+		SmartTests: []string{
+			"tests/fixtures/smart-tests-git/backend/test.star",
+		},
+	}
+
+	// Find test files
+	testFiles, err := FindTestFiles("tests/fixtures/smart-tests-git", cfg)
+	if err != nil {
+		t.Fatalf("FindTestFiles failed: %v", err)
+	}
+
+	// Create a map of actual test files for easier lookup
+	actualFiles := make(map[string]struct{})
+	for _, tf := range testFiles {
+		t.Logf("Found file: %s", tf.Path)
+		actualFiles[tf.Path] = struct{}{}
+	}
+
+	// We expect to find only the directly specified file
+	expectedFiles := []string{
+		"tests/fixtures/smart-tests-git/backend/test.star",
+	}
+
+	// Verify the expected file is found
+	for _, expectedPath := range expectedFiles {
+		if _, exists := actualFiles[expectedPath]; !exists {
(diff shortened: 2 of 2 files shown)


go.mod
```diff
@@ -1,10 +1,11 @@
 module github.com/signadot/cli
 
-go 1.23
+go 1.23.0
```
scott-cotton, line 3: If you keep this as 1.23 then it allows us to upgrade minor versions of Go more easily.

internal/repoconfig/labels.go
```diff
@@ -0,0 +1,101 @@
...
+
+	labels := make(map[string]string)
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
+		if line == "" || strings.HasPrefix(line, "#") {
+			continue
+		}
+
+		parts := strings.SplitN(line, "=", 2)
```
scott-cotton, line 31: would suggest https://pkg.go.dev/strings#Cut

internal/config/test.go
```diff
@@ -45,8 +45,7 @@ type TestRun struct {
 }
 
 func (cfg *TestRun) AddFlags(cmd *cobra.Command) {
-	cmd.Flags().StringVarP(&cfg.Cluster, "cluster", "c", "", "cluster name (required)")
-	cmd.MarkFlagRequired("cluster")
+	cmd.Flags().StringVarP(&cfg.Cluster, "cluster", "c", "", "cluster name (required for test execution)")
```
scott-cotton, line 48: Looks like this will break e2es.

internal/repoconfig/config.go
```diff
@@ -0,0 +1,187 @@
...
+// LoadConfig reads the .signadot/config.yaml file from the git repository root
+func LoadConfig(startPath string) (*Config, error) {
+	// Find git repository root
+	repoRoot, err := findGitRepoRoot(startPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find git repository root: %w", err)
+	}
+
+	configPath := filepath.Join(repoRoot, ".signadot", "config.yaml")
+	fmt.Printf("Reading config from: %s\n", configPath)
```
scott-cotton, line 68: debug print

internal/repoconfig/config.go
```diff
@@ -0,0 +1,187 @@
...
+	dirLabelsCache, err := buildLabelsCache(repoRoot)
+	if err != nil {
+		return nil, err
+	}
+
+	// Use a map to deduplicate files by their full path
+	testFileMap := make(map[string]TestFile)
+
+	// Sort paths by length to process parent directories first
+	sortedPaths := sortPathsByLength(cfg.SmartTests)
```
scott-cotton, line 117: Probably best to clean the paths from the config file to avoid noise like ../dir/.. and ././ etc

internal/repoconfig/config.go
```diff
@@ -0,0 +1,187 @@
...
+	sorted := make([]string, len(paths))
+	copy(sorted, paths)
+	sort.Slice(sorted, func(i, j int) bool {
+		return len(sorted[i]) < len(sorted[j])
+	})
+	return sorted
+}
+
+// FindTestFiles finds all test files in the tests directories
+func FindTestFiles(startPath string, cfg *Config) ([]TestFile, error) {
```
scott-cotton, line 97: This function is long and has many nested levels of loops/funcs. Can we break it down some more?

internal/config/test.go
scott-cotton: never mind, didn't read it closely enough

go.mod
foxish: go mod tidy is doing this looks like - is there a way you're aware of that prevents this?

internal/repoconfig/labels.go
foxish: done, thanks!

internal/repoconfig/config.go
foxish: Modularized it a bit - hopefully it's more maintainable now.

foxish:
Fixed the CR comments and tried to break things down further. Also added some support for directly specifying the file path in config (as opposed to directory) and corresponding test. PTAL

daniel-de-vera:
These changes LGTM and will work on top of this.
@scott-cotton, given you were the initial reviewer, I wait for you to add the rubber stamp.

go.mod
scott-cotton: arg they changed it so it overwrites that now by default.

I'm ok with removing this discussion from this PR, but I think we need to either edit the go.mod manually after go mod tidy or explicitly track minor version upgrades.

PR #164, opened 2025-04-03
add --set-label to t run and change wait to --no-wait

scott-cotton:
This PR changes --wait to --no-wait for signadot test and adds a `--set-label` argument


daniel-de-vera (approved):
LGTM

PR #173, opened 2025-04-08
truncate the output of wait

scott-cotton:
without this, there is greater risk of going up the wrong number of lines to overwrite each entry with yacspin


internal/command/smarttest/printers.go
```diff
@@ -149,7 +149,7 @@ func printTestExecutionsTable(w io.Writer, txs []*models.TestexecutionsQueryResu
 		tab.AddRow(testExecRow{
 			ID:          tx.ID,
 			Source:      source,
-			TestName:    truncateTestName(testName),
+			TestName:    truncateTestName(testName, 48),
```
daniel-de-vera, line 152: Is this case problematic too?

daniel-de-vera (approved):
LGTM, just left one question

PR #177, opened 2025-04-10
chore: Fix broken usage guide link in README.md

joehorsnell:
The current link redirects to https://www.signadot.com/docs/cli, which gives a 404.


foxish (approved):
Thanks for the fix @joehorsnell!

PR #180, opened 2025-05-02
use initial user env for sandbox manager

scott-cotton:
depends on https://github.com/signadot/libconnect/pull/97


daniel-de-vera (approved):
LGTM

PR #182, opened 2025-05-05
Fix bearer-token auth in controlplaneproxy

daniel-de-vera:
Depends on https://github.com/signadot/signadot/pull/5754, https://github.com/signadot/libconnect/pull/98.

Fix `bearer-token` auth when using `controlplaneproxy`.


scott-cotton (approved):
LGTM

PR #186, opened 2025-05-29
Run go mod tidy

foxish:
Release was failing without this in https://github.com/signadot/cli/actions/runs/15331484539/job/43139095729
Just running `go mod tidy` here.


scott-cotton (approved):
LGTM

PR #188, opened 2025-06-13
addition of local env subcommands

scott-cotton:
This PR adds sandbox local environment subcommands `get-env` and `get-files`.  It is a draft PR because it is still missing 
- some handling of resources
- default output reporting of get-files

additionally,

resources are incomplete -- there's just a start on the sandbox manager client side.  As the operator doesn't yet support this, this code is dead in this PR but helps prepare for when the operator will support it.


daniel-de-vera (approved):
LGTM - modulo the files GC which is still under discussion and can be implemented separately.

PR #189, opened 2025-06-16
signadot sb get-{env,files}: addition of valueFrom: resource

scott-cotton:
This PR adds support for `valueFrom: resource` in sandbox local env and files.

Stacked on #188 

depends on https://github.com/signadot/go-sdk/pull/68
and https://github.com/signadot/libconnect/pull/101


daniel-de-vera:
> > @scott-cotton, check this case:
> 
> a consequence of a dyslexia-infused codebase. See [0b34fd7](https://github.com/signadot/cli/commit/0b34fd7423776769b2acd2c7487cbef549107f06)

And any reason why doing the fix here instead of within `tunnel-api`?

daniel-de-vera:
> If you feel strongly about it I can move it to the tunnel-api but would prefer to do so before the next xrc in that case.

I don’t feel strongly about it, but I do have a slight preference for it, it feels a bit more natural and easier to consume.
But I leave it up to you. Let me know what you will do so I know if I go ahead and cut the xrc or wait for this change.

scott-cotton:
> > If you feel strongly about it I can move it to the tunnel-api but would prefer to do so before the next xrc in that case.
> 
> I don’t feel strongly about it, but I do have a slight preference for it, it feels a bit more natural and easier to consume. But I leave it up to you. Let me know what you will do so I know if I go ahead and cut the xrc or wait for this change.

I have a tunnel-api PR here: https://github.com/signadot/signadot/pull/5910

I need to run for a while.  If all looks good to you there, please feel free to merge to unblock the xrc and then I'll undo the reversals here when I get back.

daniel-de-vera (approved):
LGTM

PR #194, opened 2025-07-03
add downward API to get-env output

scott-cotton:
easy improvement from request from Arjun for demo: add downward api


daniel-de-vera (approved):
LGTM

arjun-sd:
There's the issue of how to deal with the conflict when you have multiple local sandboxes running but this issue exists independent of downward api.

PR #208, opened 2025-09-09
This PR fixes a bug with slow fork-exec

scott-cotton:
Some tools, such as https://www.airlockdigital.com interfere with fork/exec pattern in that there is an increased delay between `cmd.Start` and when that command actually starts running.

In our fork/exec chain for `signadot local`, after `cmd.Start()` we check if the sub-process is ready in the sense of writing its pid file.

This PR
1. increases the timeout for this from 1 to 30sec; and
2. does not return an error if there is a timeout waiting for the sub-process to be ready because that causes `signadot local connect` to report an error but not cleanup, but the background process is still there.  Rather it sends a warning to the log.
3. improves error reporting (depends in part on https://github.com/signadot/libconnect/pull/125)

### DIFF
--- a/internal/command/locald/locald.go
+++ b/internal/command/locald/locald.go
@@ -39,12 +39,12 @@ func run(cfg *config.LocalDaemon, args []string) error {
 	signal.Ignore(syscall.SIGHUP)
 
 	if err := cfg.InitLocalDaemon(); err != nil {
-		return err
+		return fmt.Errorf("could not init local daemon config: %w", err)
 	}
 	ciConfig := cfg.ConnectInvocationConfig
 	log, err := getLogger(ciConfig, cfg.RootManager)
 	if err != nil {
-		return err
+		return fmt.Errorf("could not get logger: %w", err)
 	}
 	pidFile := ciConfig.GetPIDfile(cfg.RootManager)
 
@@ -53,7 +53,7 @@ func run(cfg *config.LocalDaemon, args []string) error {
 		// it is is independent of PATH.
 		binary, err := exec.LookPath(os.Args[0])
 		if err != nil {
-			return err
+			return fmt.Errorf("locald --daemon: error finding executable path for %s: %w", os.Args[0], err)
 		}
 		args := []string{"locald"}
 		env := []string{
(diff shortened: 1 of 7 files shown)


internal/locald/rootmanager/root_manager.go
```diff
@@ -102,9 +102,10 @@ func (m *rootManager) Run(ctx context.Context) error {
 }
 
 func (m *rootManager) runAPIServer(ctx context.Context) error {
-	ln, err := net.Listen("tcp", fmt.Sprintf(":%d", m.ciConfig.LocalNetPort))
+	addr := fmt.Sprintf(":%d", m.ciConfig.LocalNetPort)
+	ln, err := net.Listen("tcp", addr)
 	if err != nil {
-		return err
+		return fmt.Errorf("error running api server: couldnot listen on %s: %w", addr, err)
```
joehorsnell, line 108: Nit - typo should be either `couldn't` or `could not`

internal/command/locald/locald.go
```diff
@@ -75,15 +75,21 @@ func run(cfg *config.LocalDaemon, args []string) error {
 		cmd.Env = env
 
 		if err := cmd.Start(); err != nil {
-			return err
+			return fmt.Errorf("locald %s --daemon: error starting command binary=%s args=%v: %w", mgr(cfg), binary, cmd.Args, err)
 		}
 		// and check to see that it succeeded
-		return processes.WaitReady(pidFile, time.Second, cmd.Process, log)
+		if err := processes.WaitReady(pidFile, 30*time.Second, cmd.Process, log); err != nil {
```
joehorsnell, line 81: This seems to be the key line - nice 👌

My only thought here is that going from one second to 30 is perhaps a bit too much? In the error case (eg if AirLock blocks process execution completely), will this then hang for 30s before terminating?

You could maybe increase it to 10 seconds by default, and potentially provide a `--timeout` command-line switch to the `connect` sub-command to increase it, if necessary?

joehorsnell (commented):
Looks good @scott-cotton - thanks.

A couple of comments 👍

internal/command/locald/locald.go
scott-cotton: Ideally I think one should plum the `--wait-timeout` flag from `signadot local connect` here but priorities prevent me from doing that just now.

Yes, if AirLock or something blocks it completely or beyond `--wait-timeout`, it will hang the call to `signadot local connect`. 

I'll see if I can add the `--wait-timeout` plumbing.

scott-cotton: I think the timeout there should be the same thing as `--wait-timeout` as passed to `signadot local connect` (and the --wait-timeout is already there for `signadot local connect`, and has a default of 10sec).

joehorsnell: Sounds good 👍.

Thanks again for getting this fix done quickly!

scott-cotton: This is done in https://github.com/signadot/cli/pull/208/commits/d80c1a0a0951c82ccdbf68af223e1a9736582e66

scott-cotton: > Thanks again for getting this fix done quickly!

Thanks for your help debugging it!

joehorsnell (approved):

PR #213, opened 2025-10-02
traffic watch base functionality

scott-cotton:
depends on https://github.com/signadot/libconnect/pull/131 and https://github.com/signadot/signadot/pull/6249

TODO probably on stacked PR(s):
- instrumenting and undoing sandbox `middleware:` changes
- missing: -o {yaml/json} support, 
- improve error messages 
- default output dir?
- timeout options?
- flag naming and help message improvements

But, hey, it works!  tested locally in various scenarios:

- `--meta-only` and not
- restarting tw-server
- GC'ing the connection
- setting timeout with context
- control-C

### DIFF
--- a/internal/trafficwatch/get.go
+++ b/internal/trafficwatch/get.go
@@ -180,8 +180,8 @@ func ensureDir(p string) error {
 	}
 	if err != nil {
 		err := os.Mkdir(p, 0755)
-		if err != nil && os.IsExist(err) {
-			return nil
+		if err != nil && !os.IsExist(err) {
+			return err
 		}
 	}
 	return nil


daniel-de-vera:
@scott-cotton, something that was making some noise on my mind for a while (and that I finally see it clear now) is the fact that the `--meta-only` mode doesn’t really tell you much about the request. So, from a user’s perspective, it feels kind of incomplete: I know we’re intercepting a request for a given routing key and host, but I have no visibility into what the request actually is (method, path, etc.).

Because of that, I don’t see much value in running with `--meta-only` mode as it stands. In my opinion, when running in this “lightweight” mode, user expectations would be something closer to an access log: no bodies streamed, but still enough context to view the core request/response details.

I guess it might be worth digging deeper into this aspect of the devex, since  it could imply some changes in the operator.

scott-cotton:
> @scott-cotton, something that was making some noise on my mind for a while (and that I finally see it clear now) is the fact that the `--meta-only` mode doesn’t really tell you much about the request. So, from a user’s perspective, it feels kind of incomplete: I know we’re intercepting a request for a given routing key and host, but I have no visibility into what the request actually is (method, path, etc.).
> 
> Because of that, I don’t see much value in running with `--meta-only` mode as it stands. In my opinion, when running in this “lightweight” mode, user expectations would be something closer to an access log: no bodies streamed, but still enough context to view the core request/response details.
> 
> I guess it might be worth digging deeper into this aspect of the devex, since it could imply some changes in the operator.

Indeed metadata here is metadata about the request outside of any TCP payload: it's the extra stuff we know due to the middleware/traffic-manager system.   That's clearly outside the kinds of things you mention, but also necessary and critically gives a request id and time and host.

While I agree folks are likely to want more in metadata, I disagree that --meta-only is not useful. It shows when things happen to what hosts, live.

For adding non-metadata in the sense of stuff in TCP payloads, I guess it should be possible to restrict to request headers, HTTP protocol info, and possibly the same for response headers and include info without needing to stream everything.  However, I am unsure of what you are implying w.r.t. MVP and our timeline: as this observation is quite late in that regard I'm unsure of what to do in terms of that and this comment.

scott-cotton:
I have changed the output a bit and fixed a problem that led to some missing log messages in https://github.com/signadot/cli/pull/213/commits/3f28b913eb83bbe7a7ec5f83117c5d3e86f31a09

internal/config/trafficwatch.go
```diff
@@ -0,0 +1,22 @@
...
+	ToDir       string
+	Sandbox     string
+	Short       bool
+	HeadersOnly bool
+}
+
+func (c *TrafficWatch) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&c.Sandbox, "sandbox", "", "sandbox whose traffic to watch")
+	cmd.Flags().BoolVar(&c.Short, "short", false, "only watch request metadata")
+	cmd.Flags().BoolVar(&c.HeadersOnly, "headers-only", false, "do not record request bodies")
```
daniel-de-vera: Maybe "do not record request **and response** bodies"

PR #218, opened 2025-10-08
Override - Add support for policies and logs args

davixcky:
- **add parser for 'to' port**
- **add policies validation and args**
- **add logs for call**

image

image

image

### DIFF
--- a/internal/command/local/override/command.go
+++ b/internal/command/local/override/command.go
@@ -88,18 +88,10 @@ func runOverride(out io.Writer, cfg *config.LocalOverrideCreate) error {
 	var (
 		logServer   *http.Server
 		logListener net.Listener
-		logOutChan  chan override.LogEntry
 	)
 	logPort := int64(0)
 	if !cfg.Detach {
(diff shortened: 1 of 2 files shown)


internal/builder/sandbox.go
```diff
@@ -80,7 +82,73 @@ const (
...
 )
 
-func (sb *SandboxBuilder) AddOverrideMiddleware(worklaodPort int64, toLocal string, workloadNames ...string) *SandboxBuilder {
+type MiddlewareOverrideArg struct {
+	internal func(sb *SandboxBuilder, overrideName string) *models.SandboxesArgument
+	isSet    bool
+}
+
+func NewOverrideArgPolicy(defaultFallThroughStatus string, unimplementedResponseCodes []int) (*MiddlewareOverrideArg, error) {
+	policies := make(map[string]interface{})
```
scott-cotton, line 91: let's use the newer and more readable `any` instead of `interface{}` (they are the same thing just different names)

PR #223, opened 2025-10-16
delay doneAt as needed

scott-cotton:
This adds a filter to doneAt that delays it until after the initial received meta has been logged

### DIFF
--- a/internal/trafficwatch/mw_request_done.go
+++ b/internal/trafficwatch/mw_request_done.go
@@ -13,7 +13,7 @@ import (
 func waitLogged(logged, done <-chan string) chan string {
 	res := make(chan string)
 	go func() {
-		d := map[string]bool{}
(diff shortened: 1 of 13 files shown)


internal/trafficwatch/mw_request_done.go
```diff
@@ -10,6 +10,37 @@ import (
...
+					done = nil
+				} else {
+					res <- id
+					delete(d, id)
+				}
+			}
+		}
+	}()
+	return res
+}
```
daniel-de-vera: Are we supposed to use `d` within this code?

scott-cotton: fixed but in a merge commit due to merge conflicts
https://github.com/signadot/cli/pull/223/commits/6fd0371e45264db427fe1074594de5b4ae93bb4f

PR #226, opened 2025-10-21
Skeleton TUI

davixcky:
- Adds dependencies
- Basic layout
- Mock data
- Custom types for HttpRequest and HttpResponse (With the current format that Scott is returning i can construct this using the stdlib)
image
image
image
image
image

### DIFF
--- a/internal/command/traffic/command.go
+++ b/internal/command/traffic/command.go
@@ -17,6 +17,7 @@ func New(api *config.API) *cobra.Command {
 	// Subcommands
 	cmd.AddCommand(
 		newRecord(cfg),
+		newInspect(cfg),
 	)
 
 	return cmd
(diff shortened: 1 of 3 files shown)


internal/tui/models/http_request.go
```diff
@@ -0,0 +1,97 @@
...
+	Duration   time.Duration     `json:"duration"`
+}
+
+// HTTPRequestList represents a list of HTTP requests
+type HTTPRequestList struct {
+	Requests []HTTPRequest `json:"requests"`
+	Total    int           `json:"total"`
+	Page     int           `json:"page"`
+	PerPage  int           `json:"per_page"`
+}
```
daniel-de-vera: I know the data is still mocked at this stage, but this structure makes me think we'll end up keeping the full request and response bodies in memory for a set of requests. That might not be ideal.

Wouldn't it be better to load the bodies on demand, for example, only when the user navigates to the right panel and opens the request/response details?

daniel-de-vera (approved):
Approving this, in case it's easier to addres my last comment in the subsequent PR.

Issue #228, opened 2025-10-22
[CLI] Override exiting with 401 mid session


foxish:
```
(localhost:8000) GET     /api/v1/vars/latest-operator-version -> 200
exiting due to problem with sandbox: 401 Unauthorized: unauthorized
→ Removing override middleware in dev-apisvr-foxish
```

But it doesn't actually seem like it's unauthorized becaue I can make the command work when I restart it - perhaps this is an effect of overriding the apiserver partly?

cc @daniel-de-vera

daniel-de-vera:
FYI: https://github.com/signadot/cli/pull/231 fixes this issue.

foxish:
Haven't seen the issue since, so closing. Thanks!

PR #231, opened 2025-10-23
Context handling improvements and auth fixes

daniel-de-vera:
**This PR includes the following changes**:

- Improves context handling across various CLI commands by properly linking contexts to signals, making undo operations more reliable.
- Fixes the auth token refresh issue reported in https://github.com/signadot/cli/issues/228.
- Normalizes and aligns code between the `override` and `traffic record` commands.

Brief demo (has audio):

[screen-capture (43).webm](https://github.com/user-attachments/assets/f2beb0aa-e6c3-4613-aa8d-012a10134175)

### DIFF
--- a/internal/command/local/override/command.go
+++ b/internal/command/local/override/command.go
@@ -1,0 +1,63 @@
+package override
+
+import (
+	"github.com/signadot/cli/internal/config"
+	"github.com/spf13/cobra"
+)
+
+func New(local *config.Local) *cobra.Command {
+	cfg := &config.LocalOverrideCreate{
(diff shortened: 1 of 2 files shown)


daniel-de-vera:
> > > But maybe some of those context.Background calls on cleanup should somehow take the overall command invocation context instead.
> > > Unless the extra plumbing complexity doesn't seem worth it...
> > 
> > 
> > The problem is that in some cases (e.g. on ctr-c), the context may be already cancelled.
> 
> I think the problem is also that there is a containing context whose cancelation is not respected. One could have a context that does not cancel on control-C as the root, for example.

Ok, let me see what I can do.

PR #233, opened 2025-10-24
traffic record: bugfix for attempting to apply deprecated fields

scott-cotton:
our document style flow is quite broken.


internal/command/traffic/instrument.go
```diff
@@ -72,6 +72,11 @@ func applyWithLocal(cfg *config.TrafficWatch, sb *models.Sandbox) error {
 		}
 		sb.Spec.LocalMachineID = machineID
 	}
+	// remove deprecated
+	sb.Spec.Endpoints = nil
+	for _, f := range sb.Spec.Forks {
+		f.Endpoints = nil
+	}
```
daniel-de-vera: Don't we need the same in override?

PR #236, opened 2025-10-28
add basic error handling for filemanager

davixcky:
This PR allows to differentiate in the TUI if there is an error without panic

image
image

### DIFF
--- a/internal/tui/filemanager/config.go
+++ b/internal/tui/filemanager/config.go
@@ -5,6 +5,7 @@ import (
 	"path/filepath"
 
 	"github.com/signadot/cli/internal/config"
+	"github.com/signadot/libconnect/common/trafficwatch/api"
 )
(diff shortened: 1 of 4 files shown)


internal/tui/filemanager/config.go
```diff
@@ -5,10 +5,21 @@ import (
...
+type MessageType string
+
+const (
+	MessageTypeData            MessageType = "data"
+	MessageTypeStatusNoStarted MessageType = "target_not_found"
+)
+
+type LineMessage struct {
+	MessageType MessageType
+	Data        any
```
scott-cotton: I think we should

1) have another field for the error, say `Error error`.
2) make Data a string

The tagged union pattern will work fine here.

scott-cotton (approved):
LGTM

PR #237, opened 2025-10-28
Add logs support

davixcky:
This adds a scanner similar to the tui parser of the requests. However given that i introduced some changes in #236 i didn't use that one. The idea is to create like a base scanner and then make it so can receive some functions for parsing and etc.

image

### DIFF
--- a/internal/tui/components/status.go
+++ b/internal/tui/components/status.go
@@ -1,7 +1,6 @@
 package components
 
 import (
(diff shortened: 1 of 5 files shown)


daniel-de-vera (approved):
LGTM: approving this to unblock further development, having into account that @scott-cotton will refactor this in a separate PR.

PR #238, opened 2025-10-28
[TUI] - Improve UI/UX traffic record

davixcky:
image

image

image


foxish:
> Please add timestamps - they're necessary for basic usability IMO.

I'd propose the following for the left hand side pane.
```
<Timestamp> <HTTP/gRPC> Method (if HTTP) URI (host/path)
```

foxish (approved):
LGTM, as per our sync via DM assuming that the pending issues will be resolved in future PRs

PR #242, opened 2025-10-30
[TUI] - Fix left pane UI rendering

davixcky:
image


foxish:
When I send requests to the frontend, I think something is throwing off the height of the TUI.
It seems to grow the left pane so much and the right pane is fixed

Screenshot 2025-10-30 at 9 17 19 AM

foxish (approved):
LGTM for the line format but there's an issue in https://github.com/signadot/cli/pull/242#issuecomment-3468843637 that you can address here or subsequently.

PR #247, opened 2025-10-30
[TUI] Revert rendering order and add follow mode

davixcky:
- Add follow mode
- Revert to the original ordering
- Move messages to file
- Make sure paginator is always at the bottom

image
image
image

### DIFF
--- a/internal/tui/components/keys.go
+++ b/internal/tui/components/keys.go
@@ -95,15 +95,11 @@ func (k KeyMap) ShortHelp() []key.Binding {
 // FullHelp returns keybindings for the expanded help view. It's part of the
 // key.Map interface.
 func (k KeyMap) FullHelp() [][]key.Binding {
-
-	tmpFollowMode := k.FollowMode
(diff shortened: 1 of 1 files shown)


internal/tui/components/keys.go
```diff
@@ -91,11 +95,15 @@ func (k KeyMap) ShortHelp() []key.Binding {
 // FullHelp returns keybindings for the expanded help view. It's part of the
 // key.Map interface.
 func (k KeyMap) FullHelp() [][]key.Binding {
+
+	tmpFollowMode := k.FollowMode
+	tmpFollowMode.SetHelp("f", "turn on/off follow mode")
```
davixcky: Remove this, not needed anymore

foxish: Also, we can just say "toggle follow mode" instead of "on/off" - wherever we show the help text.

daniel-de-vera:
@davixcky, the right panel is not getting refreshed on the follow mode (when a new request comes in)

Issue #243, opened 2025-10-30
[Override] Bug in printed info with override


foxish:
```
signadot local override --sandbox=pr-263-location --workload=location --port=8083 --to=8083
→ Applying override to pr-263-location
Waiting (up to --wait-timeout=3m0s) for sandbox to be ready...
✓ Sandbox status: Ready: All desired workloads are available.
✓ Override created port 8083 to localhost:8083

Override has been set up successfully.
* If localhost:8083 responds with `sd-override: true` header
    -> Final response from: localhost:8083
* Default: (All other local responses)
    -> Final response from: location:8083 (Sandbox)
```

The default case should say the sandbox name and not localhost:8083

cc @daniel-de-vera

PR #252, opened 2025-10-31
Fix override text printed once command is started

foxish:
Fixes https://github.com/signadot/cli/issues/243

Case 1:
Screenshot 2025-10-31 at 9 21 14 AM

Case 2:
Screenshot 2025-10-31 at 9 21 48 AM


scott-cotton (approved):
LGTM

PR #253, opened 2025-10-31
Amend override long description

foxish:
Same content, just organized it a bit for easier comprehension.

### DIFF
--- a/internal/command/local/override/command.go
+++ b/internal/command/local/override/command.go
@@ -19,7 +19,7 @@ How it works:
 - Any HTTP request (including gRPC) to your sandbox is first routed to your local service.
 - Your local service processes the request and replies.
 - By default:
-    * If your local service's response contains the header 'sd-override: true' (for HTTP), or the metadata 'sd-override: true' (for gRPC), the response from your local service is immediately sent back to the client—the request does NOT reach the original sandbox workload.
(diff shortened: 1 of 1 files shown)


internal/command/local/override/command.go
```diff
@@ -13,24 +13,22 @@ func New(local *config.Local) *cobra.Command {
...
-should not be overridden. When set, all other traffic will be overridden
-except for the specified status codes, which will fall through to the
-original sandboxed destination.`,
+		Long: `The 'override' command lets you intercept both HTTP and gRPC traffic coming into your sandbox and process it with a local service you specify (such as on your laptop).
+
+How it works:
+- Any HTTP or gRPC request to your sandbox is first routed to your local service.
+- Your local service processes the request and replies.
+- By default:
+    * If your local service's response contains the header 'sd-override: true' (for HTTP), or the metadata 'sd-override: true' (for gRPC), the response from your local service is immediately sent back to the client—the request does NOT reach the original sandbox workload.
```
scott-cotton: > or the metadata 'sd-override: true' (for gRPC)

I have not tested this.  Looked at the implementation, and it looks like it would likely work, but it would go through several implicit coding steps under the hood:  gRPC metadata -> http2 headers -> plain Go http headers (using http2 transport on top of http).

I think @daniel-de-vera tested this?

foxish: I removed the reference to metadata, @daniel-de-vera  tells me it's untested and we can leave it as "set a response header" instead for both HTTP and gRPC.

PR #259, opened 2025-10-31
tui improve help view

davixcky:
- **make help view more colored and organized**
- **fix left pane not showing request on follow mode but focus right**


foxish:
Seems not responsive @davixcky 

image

PR #268, opened 2025-11-04
improvements to tw reliability

scott-cotton:
uses new libconnect with improvements for reliability 
(depends on https://github.com/signadot/libconnect/pull/142)

However, with those changes, the buffering introduced here is not actually necessary, just better.

note also has a bug fix for a file leak...


daniel-de-vera (approved):
LGTM

PR #269, opened 2025-11-04
fix default dir for traffic inspect

scott-cotton:


internal/command/traffic/inspect.go
```diff
@@ -41,6 +44,16 @@ becomes available in the specified directory.`,
...
+	if cfg.Directory == "" {
+		signadotDir, err := system.GetSignadotDir()
+		if err != nil {
+			return err
+		}
+		// default to traffic record default with json
+		dirSuffix := "-json"
+		relDir := trafficwatch.DefaultDirRelative + dirSuffix
+		cfg.Directory = filepath.Join(signadotDir, relDir)
+	}
```
daniel-de-vera: We have a similar logic in:

https://github.com/signadot/cli/blob/e36e5e698c914ea644c5f3368fe9931a81851705/internal/command/traffic/record.go#L101-L110

Maybe it would be nice to place that into a common function?

scott-cotton: I have refactored https://github.com/signadot/cli/pull/269/commits/8d4af708e5aebbb9b08b950c34bfc1cfb2223956 but I'm not sure if that's any better.

PR #272, opened 2025-11-07
add support for no keyring login

davixcky:
image

Adding `--no-keyring` to allow users with no keyring to use the app. This is insecure given the token can be acessed by reading  `~/.signadot/credentials`

### DIFF
--- a/internal/config/auth.go
+++ b/internal/config/auth.go
@@ -15,7 +15,7 @@ type AuthLogin struct {
 
 func (c *AuthLogin) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&c.WithAPIKey, "with-api-key", "", "log in using the provided API key.")
-	cmd.Flags().BoolVar(&c.PlainText, "no-keyring", false, "store credentials in plain text file instead of keyring")
+	cmd.Flags().BoolVar(&c.PlainText, "insecure-storage", false, "store credentials in plain text file instead of keyring")
 }
 
 type AuthStatus struct {


scott-cotton:
@davixcky overall looks good (and necessary)

But, when I run with `--insecure-storage`,  I get this

```
25-11-12 scott@air cli % ./signadot auth login --insecure-storage
Opening browser at:
to authenticate using the code:

✗ Waiting for authentication:
Error: couldn't get device token: 400 Bad Request: deviceCode must be defined
25-11-12 scott@air cli %
```

and the finder opens up (Mac OS/X)



Also, looks to me like

internal/auth/auth.go should not fall back to plaintext unless `--insecure-storage` is supplied as an option.  Is this falling back w/out the option?  If so, can we plumb the option to auth, eg with

```
func ResolveAuth(opts ...AuthOpts)
```
?

foxish:
Wondering if another idea to consider is to do pure in-memory storage and allow it to go away once the daemon is terminated. Wdyt?

EDIT: just realized we don't always have the daemon running because local connect may not be run, never mind.

davixcky:
> 25-11-12 scott@air cli % ./signadot auth login --insecure-storage
Opening browser at:
to authenticate using the code:
>
> ✗ Waiting for authentication:
Error: couldn't get device token: 400 Bad Request: deviceCode must be defined
25-11-12 scott@air cli %

Are you testing against staging or local api?


> and the finder opens up (Mac OS/X)

That's weird, i haven't been able to reproduce. This happens time to time?

> internal/auth/auth.go should not fall back to plaintext unless --insecure-storage is supplied as an option. Is this falling back w/out the option? If so, can we plumb the option to auth, eg with

I don't think that's true given that otherwise will make user to use that flag everytime. I tried to mimic the GH behavior which you only specify the first time

scott-cotton:
> Are you testing against staging or local api?

> > and the finder opens up (Mac OS/X)
> That's weird, i haven't been able to reproduce. This happens time to time?

Just took a look and indeed it's only local (and also occurs w/out --insecure-storage)

scott-cotton (approved):
LGTM

PR #301, opened 2026-02-27
Add `signadot auth token` command

scott-cotton:
## Summary
- Adds \`signadot auth token\` subcommand that prints the bearer token to stdout (no newline) for scripting and piping
- Returns the cached token if still valid; refreshes transparently via the existing \`checkKeyringAuth\` flow only when expired
- Errors with a clear message when using API key auth (no bearer token involved)
- Errors when not authenticated

## Test plan
- [x] \`signadot auth token\` prints a valid bearer token after \`signadot auth login\`
- [x] \`signadot auth token\` returns the cached token without a network call when not expired
- [x] \`signadot auth token\` refreshes automatically when the token is expired (requires an expired token to exercise)
- [x] \`signadot auth token\` errors when not authenticated (requires logging out of keyring)
- [x] \`signadot auth token\` errors with a helpful message when using API key auth
- [x] Output has no trailing newline (\`signadot auth token | wc -c\`)

### DIFF
--- a/internal/command/auth/auth.go
+++ b/internal/command/auth/auth.go
@@ -18,6 +18,7 @@ func New(api *config.API) *cobra.Command {
 		newLogin(cfg),
 		newStatus(cfg),
 		newLogout(cfg),
+		newToken(cfg),
 	)
 
 	return cmd
--- a/internal/command/auth/token.go
+++ b/internal/command/auth/token.go
@@ -20,7 +20,7 @@ func newToken(cfg *config.Auth) *cobra.Command {
 }
 
 func runToken(cfg *config.Auth, out io.Writer) error {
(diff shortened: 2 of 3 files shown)


foxish (approved):
LGTM

joehorsnell:
Thanks for implementing this feature @scott-cotton and @foxish!

Just one question, out of interest - are there any tests for this CLI functionality? I was expecting to see some CLI tests to invoke the new command, check the token generation, etc.?

joehorsnell:
Related - I noticed that the [Signadot CLI auth docs](https://www.signadot.com/docs/reference/cli/auth) had not been updated, so created an [issue](https://github.com/signadot/community/issues/126) to suggest adding some @scott-cotton.

PR #315, opened 2026-04-06
Fix get-output --all --dir to include step-scoped outputs

davixcky:
## Summary
- `plan x get-output <exec-id> --all --dir` now exports both plan-level and step-level outputs
- Step-level outputs are written to `<dir>/<step-id>/<output-name>` subdirectories
- Metadata sidecars (`--metadata`) also supported for step-level outputs
- Deduplication: step outputs already promoted to plan scope are not exported twice

## Reproduce (before)

```
$ signadot plan x outputs tc78bkcfs9xdw
NAME      STEP                SCOPE   STORAGE   SIZE   READY
capture   send_http_request   step    inline    765B   true
result    check_status_code   step    inline    35B    true

$ mkdir -p /tmp/out && signadot plan x get-output tc78bkcfs9xdw --all --dir /tmp/out/
No outputs.

$ ls /tmp/out/
(empty)
```

## Reproduce (after)

```
$ mkdir -p /tmp/out && signadot plan x get-output tc78bkcfs9xdw --all --dir /tmp/out/
Exported /tmp/out/send_http_request/capture
Exported /tmp/out/check_status_code/result

$ ls -R /tmp/out/
check_status_code
send_http_request

/tmp/out/check_status_code:
result

/tmp/out/send_http_request:
capture

$ cat /tmp/out/check_status_code/result
{"check": "Verify HTTP Status Code"}
```

## Test plan
- [x] `--all --dir` exports step-scoped outputs into `<dir>/<step>/<name>` subdirectories
- [x] `--all --dir --metadata` writes `.meta.json` sidecar files for step outputs when metadata exists
- [x] Plan-level outputs still export correctly (to `<dir>/<name>`)
- [x] Step outputs already promoted to plan scope are not duplicated
- [x] "No outputs." only shown when truly no outputs exist at any scope

### DIFF
--- a/internal/command/planexec/get_output.go
+++ b/internal/command/planexec/get_output.go
@@ -13,6 +13,7 @@ import (
 	"github.com/signadot/cli/internal/config"
 	"github.com/signadot/go-sdk/client"
 	planexecs "github.com/signadot/go-sdk/client/plan_executions"
+	"github.com/signadot/go-sdk/models"
 	"github.com/spf13/cobra"
 )
 
@@ -111,11 +112,15 @@ func getAllOutputs(cfg *config.PlanExecGetOutput, log io.Writer, execID string) 
 		return err
 	}
 
-	outputs := resp.Payload.Status.Outputs
-	if len(outputs) == 0 {
+	// Reuse collectAllOutputs to gather plan-level and step-level outputs.
+	all := collectAllOutputs(resp.Payload)
+	if len(all) == 0 {
 		fmt.Fprintln(log, "No outputs.")
 		return nil
 	}
+
+	// Build a metadata lookup from the raw response for sidecar export.
+	metadataMap := buildMetadataMap(resp.Payload)
 
 	if err := os.MkdirAll(cfg.Dir, 0o755); err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


scott-cotton:
@davixcky I agree there's a ux problem here and think this helps but I also think there are some problems.

I would recommend to change the output dir layout to avoid output name conflicts by construction so that --all --dir works with step and plan output at once without conflicts in naming.  Perhaps something like

```
./plan/<output-1>
./steps/<step>/<output-1>
```

one could also possibly replace `steps` with something that cannot be an output name (any invalid expr-language identifier) and then keep the plan level outputs at the top level, maybe something like `:steps` would work (please check if you try this).

Second issue is that I think that when a plan output references a step output, we should just drop tracking the reference and output the plan level and step level independently.

davixcky:
@scott-cotton Addressed both points:

**1. Directory layout — using `:steps` prefix**

Plan outputs stay at the top level, step outputs go under `:steps/`:

```
<dir>/<output-name>                    # plan-level
<dir>/:steps/<step-id>/<output-name>   # step-level
```

Verified `:steps` works as a directory name on Linux. It cannot collide with output names since identifiers must start with a letter per `VarRx` (`^[a-zA-Z][a-zA-Z0-9_.-]*$`).

**2. Dropped reference tracking — plan and step outputs are independent**

Removed the dedup logic. When a plan output references a step output via `StepRef`, both are now exported independently.

**Test results** against execution `v8gvjcbjdv654` which has plan output `check_result` referencing step `check_status/result`:

```
$ signadot plan x outputs v8gvjcbjdv654
NAME           STEP           SCOPE   STORAGE   SIZE      READY
check_result                  plan    inline    24B       true
capture        send_request   step    inline    2.507kB   true
result         check_status   step    inline    24B       true

$ signadot plan x get-output v8gvjcbjdv654 --all --dir /tmp/out-test/
Exported /tmp/out-test/check_result
Exported /tmp/out-test/:steps/send_request/capture
Exported /tmp/out-test/:steps/check_status/result

$ find /tmp/out-test -type f | sort
/tmp/out-test/check_result
/tmp/out-test/:steps/check_status/result
/tmp/out-test/:steps/send_request/capture
```

Both `check_result` (plan) and `:steps/check_status/result` (step) are exported — no dedup. Single output downloads (`get-output <id> <name>` and `get-output <id> <step>/<name>`) also verified working.

scott-cotton (approved):
LGTM

PR #318, opened 2026-04-14
plan x get-output: sidecars by default with server Content-Type

scott-cotton:
Depends on https://github.com/signadot/signadot/pull/6884

## Summary

- Make \`--metadata\` default true for \`plan x get-output --all\` so sidecars ship with every bulk export.
- Emit a sidecar for every output, including ones with no explicit metadata — previously \`stdout\`/\`stderr\`/\`exitCode\` silently got no sidecar because \`buildMetadataMap\` only kept entries with non-nil \`Metadata\`, so the user had no way to tell what an exported file was without opening it.
- Record the server-supplied \`Content-Type\` header on the sidecar (keeps \`plans.ResolveContentType\` on the server as the single source of truth — no client-side re-derivation). Relies on the matching \`@Header\` swagger annotation on the apiserver; see signadot/signadot#6884.
- Drop \`omitempty\` on \`Size\` so 0-byte outputs show \`size: 0\` rather than an elided field (which was ambiguous with "size not reported").
- Bump go-sdk to \`170febfc\` to pick up the new \`ContentType\` field on \`GetStepOutputOK\` / \`GetPlanExecutionOutputOK\`.

## Example sidecar

\`\`\`json
{
  "name": "stdout",
  "step": "sidecar_test",
  "scope": "step",
  "type": "inline",
  "size": 0,
  "ready": true,
  "contentType": "text/plain; charset=utf-8"
}
\`\`\`

## Test plan

- [x] \`go build ./...\`
- [x] Manual against a live minikube: \`signadot plan x get-output <execID> --all --dir /tmp/out\` produces 4 outputs × 2 files (4 sidecars), sidecars contain \`contentType\` matching the server's header.
- [x] Verify \`--metadata=false\` still suppresses sidecars.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/nsf/jsondiff v0.0.0-20210926074059-1e845ec5d249
 	github.com/oklog/run v1.1.0
 	github.com/panta/machineid v1.0.2
-	github.com/signadot/go-sdk v0.3.8-0.20260410083957-d025f4d8f72c
+	github.com/signadot/go-sdk v0.3.8-0.20260414122927-170febfcff95
 	github.com/signadot/libconnect v0.1.1-0.20260224205539-f894c2d0a57e
 	github.com/spf13/cobra v1.8.1
 	github.com/spf13/viper v1.11.0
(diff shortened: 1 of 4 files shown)


daniel-de-vera (approved):
LGTM

PR #319, opened 2026-04-15
Extend plan x logs: list, download, and bulk export captured logs

daniel-de-vera:
## Summary

Unifies plan execution log access under a single `plan x logs` command. Previously the command only streamed live logs — now it also lists captured log streams, downloads captured stdout/stderr for a step, and bulk-exports all captured logs to a directory.

Also bumps **go-sdk** to pick up the new `DownloadStepLog` endpoint added server-side in signadot/signadot#6885.

## Command surface

Single command, intent driven by flags:

| Invocation | Behavior |
|---|---|
| `plan x logs EXEC_ID` | List captured log streams (table) |
| `plan x logs EXEC_ID STEP_ID` | Print captured stdout (default) |
| `plan x logs EXEC_ID STEP_ID -s stderr` | Print captured stderr |
| `plan x logs EXEC_ID -f` | Stream live aggregated logs |
| `plan x logs EXEC_ID STEP_ID -f` | Stream live step logs |
| `plan x logs EXEC_ID --all --dir ./logs/` | Bulk export captured logs |

Follows the `kubectl logs` convention: the default action is to print what's captured, `-f` follows the live stream.

## Edge-case messages

Instead of passing raw API errors through, the command pre-checks where it helps and produces actionable messages:

- `plan x logs EXEC_ID` on a **running** execution — table shows whatever has been captured so far, with a hint above: _"Execution is still running. Captured logs appear as steps complete. Use -f to stream live."_
- `plan x logs EXEC_ID -f` on a **terminal** execution — blocked before hitting the server (the runner is gone and live streaming can't work): _"execution is failed; live logs are unavailable. Omit -f to read captured logs"_
- `plan x logs EXEC_ID STEP_ID` when the requested stream isn't present — on a 404 we fetch the execution and diagnose: _"step "check_response" has no stdout log (available: stderr)"_
- Similar diagnoses for nonexistent step IDs and still-running steps.

Flag validation:
- `--all` requires `--dir`
- `--all` rejects a `STEP_ID` argument
- `--all` cannot be combined with `-f`

### Automated (unit tests, `logs_test.go`)
- `renderCapturedLogs` prints the "still running" hint when phase is `running`
- No hint when phase is `completed` / `failed` / `cancelled`
- The hint is suppressed in `-o json` output (machine-readable stays clean)
- For a running execution, completed steps' captured logs still appear in the table while still-running steps without logs are omitted

### Manual (against local apiserver)
- List captured logs — completed execution with one stderr entry → correct table
- List captured logs — execution with no logs → empty table
- Download stderr from a terminal step → inline content printed
- Download stdout when only stderr was captured → helpful "available: stderr" message
- Download for a nonexistent step → "step not found in execution" message
- Bulk export with `--all --dir` → files written to `<dir>/<step-id>/<stream>`
- All three `--all` flag-combination errors
- `-o json` listing
- `-f` on a **terminal** execution → blocked by pre-check with actionable message
- `-f` on a **running** execution → kicked off via a freshly-compiled 15-second shell plan; streams and exits cleanly with code 0 on completion
- `plan x logs EXEC_ID` against that same running execution → the "still running" hint was observed in real output (not just the unit test)

## Test plan (reviewer)

- [x] `go build ./...` passes
- [x] `go test ./internal/command/planexec/...` passes
- [x] `plan x logs --help` describes all four modes
- [x] Against a completed execution: `plan x logs <id>` lists captured streams; `plan x logs <id> <step>` prints stdout
- [x] Against a running execution: `plan x logs <id>` shows the hint; `plan x logs <id> -f` streams and exits on completion
- [x] `plan x logs <id> -f` on a terminal execution returns the pre-check error rather than a 500


scott-cotton:
# Review: signadot/cli#319 + signadot/actions#11

## 1. UX consistency — the real issue

There's a live, top-level **`signadot logs --plan EXEC_ID`** command (`internal/command/logs/command.go:63`) that still streams by default, uses `--step` instead of a positional step arg, and has no captured-log / `-f` distinction. After this PR:

| Command | Default on a plan | Follow | Step |
|---|---|---|---|
| `signadot logs --plan <id>` | streams live | (implicit, always) | `--step <id>` |
| `signadot plan x logs <id>` | **lists captured** | `-f` | positional |

Two commands do the same thing with opposite defaults and different flag conventions. This is the kind of split that bakes itself into muscle memory and doesn't come out. Options, in rough order of preference:

- **Deprecate `signadot logs --plan`** and point users at `plan x logs`. The job case is `signadot logs --job`; the plan case can be its own verb.
- Alternatively, make `signadot logs --plan` adopt the same default-list-plus-`-f` shape. But that's a breaking change for anyone scripting the old command.

Either way, the PR should pick one — shipping as-is means both exist forever.

## 2. Live-vs-not-live races

- **Pre-check → SSE gap in `-f`.** `runLogs` calls `execTerminalPhase` (`logs.go:86`), gets "running", then subscribes to SSE. If termination happens in between, the SSE either returns tail+EOF or closes immediately. Functionally OK; server-side handles cleanly. Worth confirming the SSE endpoint tails recently-terminated executions (within some window) rather than 404-ing — otherwise a race loses the last 0–2s of output.
- **404 → diagnose second-GET is lying-prone.** `diagnoseMissingLog` (`logs.go:173`) issues a second `GetPlanExecution` and categorizes based on current state. A step that was still-running at the download time may have completed by the diagnose call — user gets "has no captured logs" when a retry would succeed. Not broken; misleading. One fix: add a `; retry in a moment` suffix when the state is freshly terminal.
- **`--all --dir` on a running execution is silently partial.** `bulkExportLogs` (`logs.go:205`) snapshots step logs from `GetPlanExecution` and exports only what's in the snapshot. Steps that complete after the GET are invisible. No warning that the export is incomplete. Recommend: reuse `isTerminalPhase` and print a stderr warning when the execution is non-terminal.
- **`--all` ignores `ready: false`.** The list shows a READY column, but bulk export iterates everything and calls `DownloadStepLog` even for non-ready artifacts — server will 404 and the whole export aborts mid-stream, leaving a partial tree. Recommend: skip and warn on non-ready artifacts, or retry with backoff.

## 3. Other issues

**Bugs / defensive gaps**
- `collectAllLogs` (`logs.go:372`) panics on `ex == nil` — `renderCapturedLogs` guards `running` but still calls through. Nil-check at the top.
- `bulkExportLogs`: stepID is joined via `filepath.Join(cfg.Dir, stepID)` without sanitization (`logs.go:253`). If a server-side stepID ever contained `..` or an absolute path, export escapes `--dir`. Realistically the server doesn't emit such IDs, but this is the easy kind of defense to add (`filepath.Clean` + reject `..`).
- `cfg.Dir` is silently ignored when `--all=false` — `--dir ./foo` without `--all` does nothing and doesn't tell the user.
- On download error mid-`--all`, a zero-length or partial file is left on disk with no cleanup. Worth `os.Remove(outPath)` on failure, or writing to `outPath.tmp` + rename.

**Smaller UX nits**
- Running-hint on the list is written to **stdout** (`logs.go:126`), so `plan x logs <id> | grep step1` will also match the hint line. JSON output is guarded; text isn't. Move to `log` (stderr).
- `-f` help says `"stream live logs (SSE)"` — "(SSE)" leaks an implementation detail; drop it.
- `--stream` help: `"only used with a step ID"` is no longer accurate; it's also ignored under `--all` (both streams exported regardless). Either respect `-s` under `--all` or document that it's ignored.
- `printLogsTable` renders `Size = 0` as an empty cell rather than `0 B` or `—`. Minor.

daniel-de-vera:
@scott-cotton, feedback was implemented in the three prev commits.

scott-cotton (commented):
LGTM

PR #320, opened 2026-04-20
fix: thread LocalNetPath through ciConfig for root-manager tunneling

scott-cotton:
## Summary
- Pair with signadot/libconnect#158, which now takes a `ClientConfig.LocalNetPath` string prepended to `PATH` when libconnect execs `iptables`/`pfctl`/`route`. Fixes the `exit status 127` seen on Gitpod where the user's PATH lacks `/usr/sbin`.
- Capture `$SIGNADOT_LOCALNET_PATH` at CLI invocation with `os.LookupEnv` while the user's shell env is still visible. Unset → default `/usr/sbin:/sbin`; explicitly empty → no prepend (opt-out for NixOS etc.); set → pass through. Resolved string carried on a new `ciConfig.LocalNetPath` field and wired into `localnet.ClientConfig` in the root-manager.
- The sudo `--preserve-env` list is unchanged — the env var never needs to survive the sudo elevation because the policy decision is resolved in the CLI before the daemon is spawned.
- Bumps libconnect to the paired branch commit. **Needs a final bump once signadot/libconnect#158 merges and is tagged** before this lands on main.

## Design note
Earlier iteration of this change extended `--preserve-env` to pass `SIGNADOT_LOCALNET_PATH` through sudo and had libconnect read it via `os.LookupEnv`. The ciConfig route is better because:
- ciConfig is already the canonical "user-invocation state captured and handed to the daemon" boundary; this value fits that shape.
- Removes a tacit coupling (libconnect env-var API + CLI `--preserve-env` list staying in sync across two repos).
- libconnect stays policy-free — no env-var read, no default, no unset-vs-empty tri-state. The distinction is resolved at the CLI layer where the user's shell env is visible, then a single string flows through.

## Behavior
- Default (no env var): libconnect invokes `iptables`/`pfctl`/`route` with `PATH=/usr/sbin:/sbin:$UIDPath`. Gitpod works without the `sudo ln -s /usr/sbin/iptables /usr/bin/iptables` workaround.
- `SIGNADOT_LOCALNET_PATH=...` exported by user: that value is prepended instead of the default.
- `SIGNADOT_LOCALNET_PATH=` (empty): prepend disabled, tunnel-setup execs inherit root-manager env unmodified.
- Port-forward / kubeconfig exec auth plugins: unaffected. Sandbox-manager env is not touched.

## Test plan
- [x] `go build ./...`
- [x] `go test ./...`
- [x] `gofmt -l internal/`
- [ ] Reproduce the failure on a developer env where `iptables` at `/usr/sbin/iptables` is not on user PATH, confirm `signadot local connect` succeeds.
- [ ] Confirm port-forward + kubeconfig exec auth (e.g. EKS with `aws-iam-authenticator`, GKE with `gke-gcloud-auth-plugin`) still works.
- [ ] After signadot/libconnect#158 merges: bump libconnect to the merged commit/tag.


daniel-de-vera (approved):
LGTM

PR #323, opened 2026-04-22
Add --param-secret flag to plan run

daniel-de-vera:
## Summary
- Adds `--param-secret param-name=secret-name` to `signadot plan run`, wiring a new `Secrets map[string]string` into `PlanExecutionSpec` so plan params can be bound to org secrets at execution time (without decrypted values ever touching the plan_executions row or API responses).
- Rejects overlap between `--param` and `--param-secret` for the same key locally; other validations (unknown param, missing secret) are handled server-side.
- Bumps go-sdk to pick up `PlanExecutionSpec.Secrets`.

Related: signadot/signadot#6904

## Test plan
- [ ] `signadot plan run PLAN --param-secret api-key=my-secret` → execution created, response shows `secrets: {api-key: my-secret}` (name, not value)
- [ ] `signadot plan run PLAN --param x=1 --param-secret x=my-secret` → fails locally with `param "x" appears in both --param and --param-secret; specify only one`
- [ ] `signadot plan run PLAN --param-secret unknown=my-secret` → server rejects with 400 (key must match a declared plan param)
- [ ] `signadot plan run PLAN --param-secret api-key=nonexistent` → server rejects with 400 (secret not found)
- [ ] Action script receives decrypted value via `context/api-key`, same as a literal param

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #324, opened 2026-04-24
Add secret commands

daniel-de-vera:
## Summary
- Adds `signadot secret` (alias `secrets`) with `create`, `update`, `get`, `list` (alias `ls`), and `delete` (alias `rm`) subcommands wrapping the Secrets API from the go-sdk.
- Values can be supplied via `--value` (warns about shell history), `--value-file`, `--value-stdin`, or a flat YAML/JSON file with `-f` (with `--set var=val` expansion). Value is required on both create and update, matching server-side validation.
- `get`/`list` print metadata only (`NAME`, `DESCRIPTION`, `CREATED`, `UPDATED`) since the API never returns plaintext. `-o json|yaml` supported throughout.
- Bumps `go-sdk` to a version that includes the `secrets` client.

## Test plan
- [ ] `go build ./...` and `go vet ./...` pass (verified locally)
- [ ] `signadot secret --help` and each subcommand `--help` render correctly
- [ ] Create a secret via `--value`, `--value-file`, `--value-stdin`, and `-f` paths
- [ ] Verify `--value`, `--value-file`, `--value-stdin` are rejected together
- [ ] Verify `-f` is rejected with positional NAME and with `--value*` / `--description`
- [ ] Verify missing value produces the client-side error before the API call
- [ ] Update an existing secret (value rotation) and confirm `updatedAt` changes on `get`
- [ ] `list` shows all secrets and `delete NAME` / `delete -f FILE` remove them
- [ ] `get` / `list` never include a `value` field in JSON/YAML output


scott-cotton:
Worth adding a "## Secrets" section to `agent-skills/signadot-cli/SKILL.md` in this PR (or a tight follow-up) so agents know the command exists — currently the file has zero references to `signadot secret` (only one passing mention of "Secrets" in the sandbox-file-mounts context). Bare minimum:

- `signadot secret create NAME --value VAL` (and `--value-file`, `--value-stdin`, `-f spec.yaml --set var=val`)
- `signadot secret update NAME --value VAL`
- `signadot secret get NAME` / `list` — metadata only, plaintext never returned
- `signadot secret delete NAME`
- One-line note that secrets are bound to plan params via `signadot plan run --secret param-name=secret-name` (#323)

Otherwise the entire secrets feature is invisible to agents that load this skill, and #323's `--secret` flag has no path-to-discovery from the skill side.

daniel-de-vera:
@scott-cotton, added secrets to the skill in https://github.com/signadot/cli/pull/324/changes/03a0949eead55ae00a32e32add0b0dd6c5f2e610.

scott-cotton (approved):
LGTM

PR #325, opened 2026-04-29
Add plan action list/get commands

daniel-de-vera:
## Summary

- Wires the new plan-actions API into the CLI as `signadot plan action list` and `signadot plan action get NAME` (alias `a`), under the existing `plan` parent.
- `list` shows `NAME / MANAGED BY / ENABLED / REVISION / DESCRIPTION / UPDATED` — `MANAGED BY` distinguishes system actions from org-owned actions.
- `get` shows action header (Enabled, Description, Revision, Image, Requires), parsed Inputs/Outputs/Validations tables, and the action body. Body is rendered through `glamour` when stdout is a TTY (headings, tables, code blocks, inline code); raw markdown when piped.
- `-o json` / `-o yaml` returns the raw `PlanAction` payload at both list and get.

Builds on go-sdk `feature-plan` (`PlanActionStatus.Enabled`, `PlanActionStatus.BodyImage`).

## Test plan

- [ ] `signadot plan action list` shows all visible actions with the columns above
- [ ] `signadot plan action get check` shows header + Inputs/Outputs/Validations tables + rendered body
- [ ] `signadot plan action get k6` shows the `Image:` line populated from `Status.BodyImage`
- [ ] `signadot plan action get k6 | cat` (non-TTY) prints raw markdown
- [ ] `signadot plan action get check -o json` returns the raw `PlanAction` payload
- [ ] `signadot plan action get nonexistent` surfaces a 404 from the API


scott-cotton:
From claude: 

1. formatAny mishandles non-scalar JSON values (printers.go:222-227)
  func formatAny(v any) string {
      if v == nil { return "" }
      return fmt.Sprintf("%v", v)
  }
  For a Default of {"a": 1} you get map[a:1] in the table; for [1,2] you get [1 2]. Both are misleading and ugly.
  JSON-encoding (json.Marshal → string) gives faithful round-trip output that a reader can paste back into a plan
  spec. Easy fix.

  2. formatSchema is shallow (printers.go:230)
  Extracts only top-level type; renders any non-trivial schema (array with items, oneOf, enum, object with
  properties) as plain array/object or (custom). Acceptable for a summary column, but enum is common enough it'd
  be worth one extra branch:
  if enums, ok := m["enum"].([]any); ok && len(enums) > 0 {
      return fmt.Sprintf("enum(%d)", len(enums))
  }

  3. formatImage emits input=<name> as a bare string (printers.go:158-167)
  On a "Image:" header line, input=foo reads ambiguously — could be a literal image with = in its tag or a
  Go-flag-style key=value. Suggest (from input "%s") or <input: %s> to disambiguate.

  4. PR description ↔ code mismatch. PR says list shows "Enabled/Revision/Description/Updated columns" but the
  code adds MANAGED BY as well (printers.go:18-26). Either update the description or drop the column — both are
  defensible, but they should agree. MANAGED BY is genuinely useful to differentiate org actions from system ones,
   so I'd keep it and update the PR text.

  5. printBody silently swallows the write result (printers.go:99-106)
  fmt.Fprintln(out, body) and fmt.Fprint(out, rendered) — neither return is checked. The rest of
  printActionDetails propagates errors (tabwriter Flush, sub-printers), so a broken pipe here would silently look
  successful. Minor; checking the second Fprint and returning the error keeps the contract consistent.

daniel-de-vera:
@scott-cotton, feedback addressed in https://github.com/signadot/cli/pull/325/changes/2a031d6f66b7483c8189a8956880392c08469c2a.

scott-cotton (approved):
LGTM

PR #326, opened 2026-04-29
plan x get: surface input resolution status

daniel-de-vera:
Stacked on #325.

## Summary

- Replaces the flat STEP/PHASE/ERROR table in `plan x get` with a nested per-step layout that keeps inputs and errors visually owned by the step they belong to. Per-step inputs surface refs and secret-backed bindings; pure literal/default/unbound inputs are filtered as noise.
- Adds a plan-level `Inputs:` section in the same arrow/via form. Useful for confirming which `--param-secret` bindings actually took effect at dispatch time.
- Adds `utils.TimeAgo` for the compact \"X ago\" form (the existing `FormatTimestamp` gives the verbose absolute+relative form, too wide for table columns), and uses it in `plan x list`. Several other printers inline the same `time.Parse + timeago.NoMax` pattern; migrating those is left for a separate cleanup.

## Examples

Step ref (data flow between steps):

```
Steps:
  send_request   completed
  check_status   completed
      capture    ← steps.send_request.outputs.capture   (ref)
```

Plan-level params with literal + unbound:

```
Inputs:
  present_param    (literal)
  unbound_param    (unbound)

Steps:
  step_a   completed
  step_b   completed
```

`--param-secret` binding + step failure (error nested under the failing step):

```
Inputs:
  region  ← test-input-secret  (callerSecret)

Steps:
  echo_region   failed
      error:    extra_output \"got_region\" declares a schema but the script wrote to got_region instead of got_region.json — use the .json extension for typed outputs
```

## Test plan

- [ ] `plan x get <id>` for an exec with no params shows no `Inputs:` section, just the nested steps
- [ ] `plan x get <id>` for an exec with `--param-secret` shows the `callerSecret` row with the bound secret name
- [ ] `plan x get <id>` for an exec with a step that refs another step's output shows the ref expression text under the consuming step
- [ ] `plan x get <id>` for a failed exec shows the step error nested under the failed step
- [ ] `plan x list` still shows the compact \"X ago\" form in the CREATED column
- [ ] Step inputs with only literal/default/unbound resolutions don't add visual noise

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,7 @@ require (
 	github.com/nsf/jsondiff v0.0.0-20210926074059-1e845ec5d249
 	github.com/oklog/run v1.1.0
 	github.com/panta/machineid v1.0.2
-	github.com/signadot/go-sdk v0.3.8-0.20260505171357-a02f1a78bde4
+	github.com/signadot/go-sdk v0.3.8-0.20260505200838-3bae35b0d0d8
 	github.com/signadot/libconnect v0.1.1-0.20260424105947-336dce43da75
 	github.com/spf13/cobra v1.8.1
 	github.com/spf13/viper v1.11.0
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/charmbracelet/lipgloss v1.1.1-0.20250404203927-76690c660834
 	github.com/docker/go-units v0.5.0
 	github.com/go-git/go-git/v5 v5.18.0
-	github.com/go-openapi/runtime v0.29.4
+	github.com/go-openapi/runtime v0.29.5
 	github.com/go-openapi/strfmt v0.26.2
 	github.com/goccy/go-yaml v1.10.0
 	github.com/golang/protobuf v1.5.4
@@ -22,7 +22,7 @@ require (
 	github.com/nsf/jsondiff v0.0.0-20210926074059-1e845ec5d249
 	github.com/oklog/run v1.1.0
 	github.com/panta/machineid v1.0.2
-	github.com/signadot/go-sdk v0.3.8-0.20260429134502-a68ea488880a
+	github.com/signadot/go-sdk v0.3.8-0.20260505171357-a02f1a78bde4
 	github.com/signadot/libconnect v0.1.1-0.20260424105947-336dce43da75
 	github.com/spf13/cobra v1.8.1
 	github.com/spf13/viper v1.11.0
--- a/go.sum
+++ b/go.sum
@@ -178,8 +178,8 @@ github.com/go-openapi/jsonreference v0.21.5 h1:6uCGVXU/aNF13AQNggxfysJ+5ZcU4nEAe
 github.com/go-openapi/jsonreference v0.21.5/go.mod h1:u25Bw85sX4E2jzFodh1FOKMTZLcfifd1Q+iKKOUxExw=
 github.com/go-openapi/loads v0.23.3 h1:g5Xap1JfwKkUnZdn+S0L3SzBDpcTIYzZ5Qaag0YDkKQ=
 github.com/go-openapi/loads v0.23.3/go.mod h1:NOH07zLajXo8y55hom0omlHWDVVvCwBM/S+csCK8LqA=
-github.com/go-openapi/runtime v0.29.4 h1:k2lDxrGoSAJRdhFG2tONKMpkizY/4X1cciSdtzk4Jjo=
-github.com/go-openapi/runtime v0.29.4/go.mod h1:K0k/2raY6oqXJnZAgWJB2i/12QKrhUKpZcH4PfV9P18=
+github.com/go-openapi/runtime v0.29.5 h1:uc5+/TtqLIfDBTUxnF3uppoGMt+9DzonwUWsviINlrY=
+github.com/go-openapi/runtime v0.29.5/go.mod h1:D9IUbWccdYv+km8QwmAm90FZvDcQk47vP2Y7y5as/D8=
 github.com/go-openapi/spec v0.22.4 h1:4pxGjipMKu0FzFiu/DPwN3CTBRlVM2yLf/YTWorYfDQ=
 github.com/go-openapi/spec v0.22.4/go.mod h1:WQ6Ai0VPWMZgMT4XySjlRIE6GP1bGQOtEThn3gcWLtQ=
 github.com/go-openapi/strfmt v0.26.2 h1:ysjheCh4i1rmFEo2LanhELDNucNzfWTZhUDKgWWPaFM=
@@ -210,10 +210,10 @@ github.com/go-openapi/swag/typeutils v0.26.0 h1:2kdEwdiNWy+JJdOvu5MA2IIg2SylWAFu
 github.com/go-openapi/swag/typeutils v0.26.0/go.mod h1:oovDuIUvTrEHVMqWilQzKzV4YlSKgyZmFh7AlfABNVE=
 github.com/go-openapi/swag/yamlutils v0.26.0 h1:H7O8l/8NJJQ/oiReEN+oMpnGMyt8G0hl460nRZxhLMQ=
 github.com/go-openapi/swag/yamlutils v0.26.0/go.mod h1:1evKEGAtP37Pkwcc7EWMF0hedX0/x3Rkvei2wtG/TbU=
-github.com/go-openapi/testify/enable/yaml/v2 v2.4.2 h1:5zRca5jw7lzVREKCZVNBpysDNBjj74rBh0N2BGQbSR0=
-github.com/go-openapi/testify/enable/yaml/v2 v2.4.2/go.mod h1:XVevPw5hUXuV+5AkI1u1PeAm27EQVrhXTTCPAF85LmE=
-github.com/go-openapi/testify/v2 v2.4.2 h1:tiByHpvE9uHrrKjOszax7ZvKB7QOgizBWGBLuq0ePx4=
-github.com/go-openapi/testify/v2 v2.4.2/go.mod h1:SgsVHtfooshd0tublTtJ50FPKhujf47YRqauXXOUxfw=
+github.com/go-openapi/testify/enable/yaml/v2 v2.5.0 h1:3hZD1fwydvCx/cc1R2uYNQirHqf2s6lqpKV3FcNTURA=
+github.com/go-openapi/testify/enable/yaml/v2 v2.5.0/go.mod h1:TvDZKBH7ZbMaF3EqH2AwTvNQCmzyZq8K1agRjf1B+Nk=
+github.com/go-openapi/testify/v2 v2.5.0 h1:UOCr63aAsMIDydZbZGqo5Ev01D4eydItRbekDuZMJLw=
+github.com/go-openapi/testify/v2 v2.5.0/go.mod h1:SgsVHtfooshd0tublTtJ50FPKhujf47YRqauXXOUxfw=
 github.com/go-openapi/validate v0.25.2 h1:12NsfLAwGegqbGWr2CnvT65X/Q2USJipmJ9b7xDJZz0=
 github.com/go-openapi/validate v0.25.2/go.mod h1:Pgl1LpPPGFnZ+ys4/hTlDiRYQdI1ocKypgE+8Q8BLfY=
 github.com/go-playground/assert/v2 v2.0.1/go.mod h1:VDjEfimB/XKnb+ZQfWdccd7VUvScMdVu0Titje2rxJ4=
@@ -452,8 +452,8 @@ github.com/segmentio/encoding v0.5.4 h1:OW1VRern8Nw6ITAtwSZ7Idrl3MXCFwXHPgqESYfv
 github.com/segmentio/encoding v0.5.4/go.mod h1:HS1ZKa3kSN32ZHVZ7ZLPLXWvOVIiZtyJnO1gPH1sKt0=
 github.com/sergi/go-diff v1.3.2-0.20230802210424-5b0b94c5c0d3 h1:n661drycOFuPLCN3Uc8sB6B/s6Z4t2xvBgU1htSHuq8=
 github.com/sergi/go-diff v1.3.2-0.20230802210424-5b0b94c5c0d3/go.mod h1:A0bzQcvG0E7Rwjx0REVgAGH58e96+X0MeOfepqsbeW4=
-github.com/signadot/go-sdk v0.3.8-0.20260429134502-a68ea488880a h1:S7Yc47d3onGEGKcPUrL2gNKQGuZ2yxFyr6MxVLGYHPE=
-github.com/signadot/go-sdk v0.3.8-0.20260429134502-a68ea488880a/go.mod h1:FgM6/qeQ7IdRh+hGdqWOY0ShCXkRN2bBPsWX6bILVDI=
+github.com/signadot/go-sdk v0.3.8-0.20260505171357-a02f1a78bde4 h1:egJ0Kuj7ZAHA8td1XQPiOIB6Xaj0JsjGTp/ptCUDBqU=
+github.com/signadot/go-sdk v0.3.8-0.20260505171357-a02f1a78bde4/go.mod h1:8CfvBQ/AQ3LPruaQZoflmpBjoZTwXfhBt2OtS1eet+Q=
 github.com/signadot/libconnect v0.1.1-0.20260424105947-336dce43da75 h1:LZJrEqJeSb0CGsENOAQsvDeEO2YyMY1/ir2Nz4apvbI=
 github.com/signadot/libconnect v0.1.1-0.20260424105947-336dce43da75/go.mod h1:cAsgAummH9Q9DrLQ7+S3mqrBv/+ZYKVSEXjR/WfoUJM=
 github.com/sirupsen/logrus v1.7.0/go.mod h1:yWOB1SBYBC5VeMP7gHvWumXLIWorT60ONWic61uBYv0=
--- a/go.sum
+++ b/go.sum
@@ -454,6 +454,8 @@ github.com/sergi/go-diff v1.3.2-0.20230802210424-5b0b94c5c0d3 h1:n661drycOFuPLCN
 github.com/sergi/go-diff v1.3.2-0.20230802210424-5b0b94c5c0d3/go.mod h1:A0bzQcvG0E7Rwjx0REVgAGH58e96+X0MeOfepqsbeW4=
 github.com/signadot/go-sdk v0.3.8-0.20260505171357-a02f1a78bde4 h1:egJ0Kuj7ZAHA8td1XQPiOIB6Xaj0JsjGTp/ptCUDBqU=
 github.com/signadot/go-sdk v0.3.8-0.20260505171357-a02f1a78bde4/go.mod h1:8CfvBQ/AQ3LPruaQZoflmpBjoZTwXfhBt2OtS1eet+Q=
+github.com/signadot/go-sdk v0.3.8-0.20260505200838-3bae35b0d0d8 h1:KFayX3D9o1XaDA8JNIngOVEDc0t35Y5pERsAn3Tti8A=
+github.com/signadot/go-sdk v0.3.8-0.20260505200838-3bae35b0d0d8/go.mod h1:8CfvBQ/AQ3LPruaQZoflmpBjoZTwXfhBt2OtS1eet+Q=
 github.com/signadot/libconnect v0.1.1-0.20260424105947-336dce43da75 h1:LZJrEqJeSb0CGsENOAQsvDeEO2YyMY1/ir2Nz4apvbI=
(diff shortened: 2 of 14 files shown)


scott-cotton:
**`isInterestingInput` filter is silent — make the elision explicit**

The filter at `printers.go:91-99` drops literal/default/unbound step inputs without any marker, so the rendered output silently disagrees with the underlying data. Once a user notices something was hidden, every future `plan x get` becomes "what got dropped this time?" and the trust hit compounds. The fix doesn't have to be "show everything" — explicit elision works fine.

Two cheap shapes:

**Counted summary** (one line under each step header):
```
send_request   completed
    inputs: 5 (2 ref, 3 literal)
    capture    ← steps.send_request.outputs.capture   (ref)
```

**Trailing marker** (keep the filter, name what it dropped):
```
send_request   completed
    capture    ← steps.send_request.outputs.capture   (ref)
    … 3 literal inputs hidden (use -o yaml for all)
```

Either makes the filter part of the contract instead of a hidden behavior the user has to reverse-engineer.

scott-cotton:
**`unbound` probably shouldn't be filtered at step level**

At plan level, `unbound` means the executor didn't supply a required param — surfaced via the top-level `Inputs:` section, fine. At step level, `unbound` can also mean a `ref` failed to resolve (upstream step skipped, output missing) — exactly the kind of thing `plan x get` should make obvious, not filter as noise. Suggest dropping `Unbound` from the false branch of `isInterestingInput`.

daniel-de-vera:
@scott-cotton, I added further changes, see below:

## Updates since the initial review

The original PR was scoped to making input *resolution methods* visible. After review iterations it grew into surfacing **resolved values** end-to-end, plus a fuller `plan get` view, integrating the new `PlanStepAction.Name` from the latest go-sdk, normalising the rendered output sections, and bringing `plan run --attach` to parity. Summary of the changes layered on top of the first commit:

### 1. Show every step input, with values

Earlier the printer filtered "noisy" inputs (literal/default/unbound) at the step level. That had two problems flagged in review:

- The filter was silent — the rendered output disagreed with the underlying data without any marker.
- At step level, `unbound` can mean a ref failed to resolve (upstream step skipped, output missing) — exactly the signal `plan x get` should make obvious.

Resolution: drop the filter, fetch the compiled plan via a best-effort `GetPlan`, and use it to resolve concrete values:

- Plan-level literal: from `execution.spec.params[name]`
- Step-level literal: from `step.args.values[name]` (plan author hardcoded it)
- Default: from `step.action.params[i].default` or `plan.spec.params[i].default`
- Ref: from the existing `inputs[].ref` field

If the plan can't be fetched (deleted, network blip), the printer renders without values — gracefully degrades.

### 2. Replace the ambiguous "literal" enum with prose

Both plan-level and step-level had a `literal` resolution, but they meant different things — caller passed `--param` vs plan author hardcoded. Same for `default` (plan default vs action default). The CLI now uses sentence-style source tags:

| Enum             | Plan-level prose          | Step-level prose |
|------------------|---------------------------|------------------|
| `literal`        | `(from --param)`          | `(set in plan)`  |
| `callerSecret`   | `(from --param-secret)`   | n/a              |
| `ref`            | n/a                       | `(ref)`          |
| `default`        | `(plan default)`          | `(default)`      |
| `unbound`        | `(unbound)`               | `(unbound)`      |

### 3. Visual hierarchy under each step

`error:`, `action:`, `inputs:`, and `outputs:` now sit at indent 4 (step fields), with input/output rows nested at indent 6 under their respective subheader. Previously `error:` and individual inputs shared the same indent and read like siblings.

### 4. Rebuild `plan get`

Was a thin metadata blurb (counts, names) — now a full spec view: per-param defaults/required/optional, plan-level outputs (refs, no redundant tag since they're always refs), per-step structure (action name + ID + revision, image, timeout, condition, wired inputs, declared extra_outputs).

### 5. Surface `PlanStepAction.Name` (latest SDK)

The newest go-sdk drop adds `Name` on `PlanStepAction`, plus removes `Requires` / `Validations` from `PlanActionStatus` and `Requires` from `PlanSpec`. Both `plan get` and `plan x get` now show the action label as `name (id) (revision N)`; falls back to the bare ID for plans compiled before the field landed. Plan-action and plan-spec output drops the now-empty `Requires:` line and the per-runner-group validations table.

### 6. Unified `Outputs:` section + value truncation + section normalisation

- Outputs were rendered in two ad-hoc shapes that didn't match the rest of the view: inline values as `--- name ---\n<value>` blocks and artifacts as a separate STEP/TYPE/SIZE/READY table at the end. Both folded into a single arrow-form `Outputs:` section that mirrors `Inputs:` and `Steps:`. Step-level outputs (`PlanStepStatus.Outputs`) get the same treatment inside each step block, between inputs and the next step. Together the two views show what each step emitted and how each plan-level output was wired through to the caller. Artifact tags include `ready/not ready` (always), Content-Type from `metadata.contentType` when present, and any `artifact.error`.
- Multi-line literal values broke the inline arrow form. A 38-line Playwright script as a step input rendered with the `(set in plan)` tag landing many lines below the row. `truncateForDisplay` collapses to a single line + `(N lines)` marker; over-long single-line values get a trailing ellipsis at 80 chars. Same change applies to `plan get`'s `formatValue`.
- Normalised section labels and order across both views: `Inputs → Steps → Outputs`. `plan get` previously used `Params / Outputs / Steps`; now it matches `plan x get` and `plan action get`. The reader sees the same skeleton in all three views; the via tags carry the spec-vs-runtime distinction.
- Unified the three exec-detail entry points: outputs now live in `printExecDetails`, so `plan run`, `plan x get`, and `plan x cancel` all render the same trailing `Outputs:` block. `PrintRunResult` collapses to a thin wrapper.

### 7. `plan run --attach`: surface artifact metadata, stream lifecycle uniformly

Two visual breaks fixed in the attach event stream:

- Artifact outputs rendered as `value=<nil>` because `AttachEvent` only carried Name+Value. The full `PlanOutputStatus` is in hand at emit time (the events come from a `GetPlanExecution` snapshot at terminal phase). Plumb the artifact-side fields through: `kind` (`inline` | `artifact`), `size`, `ready`, `contentType`, plus `step` for plan-level outputs that came from a step. All `omitempty` so older JSON consumers ignoring unknown fields keep working. Format parity with the rendered `Outputs:` section.
- The "Created execution X for plan Y" line was a plain stderr banner before the structured stream. Moved into the attach stream as a `type=created` event with `id` and `plan_id`, so the lifecycle reads consistently from creation through to result. Non-attach mode still prints the banner to stderr; only `--attach` switches to structured.

---

## Examples

### `plan x get` — multi-step plan with refs (action names from new SDK)

```
Inputs:
  ...

Steps:
  send_request   completed
    action: request-http (f3ppjxbzpddry)   (revision 1)
    inputs:
      url                ← http://httpbin.org/uuid   (set in plan)
      method             ← GET   (set in plan)
      headers            (unbound)
      body               (unbound)
      timeout            ← 30s   (default)
      follow_redirects   ← true   (default)
      insecure           ← false   (default)
    outputs:
      capture   ← {"request":{"host":"httpbin.org",…   (inline)

  check_status   completed
    action: check (cn20hsyg8yjmb)   (revision 1)
    inputs:
      expression     ← capture.response.statusCode == 200   (set in plan)
      name           ← Status check   (set in plan)
      attrs          (unbound)
      results_file   (unbound)
      capture        ← steps.send_request.outputs.capture   (ref)
    outputs:
      result   ← {"check":"Status check"}   (inline)

Outputs:
  check_result   ← {"check":"Status check"}   (inline, from check_status)
```

### `plan x get` — `--param-secret` binding + step failure

```
Inputs:
  region   ← test-input-secret   (from --param-secret)

Steps:
  echo_region   failed
    action: run-image (3xpvgr1f8wfbc)   (revision 1)
    error: extra_output "got_region" declares a schema but the script wrote to got_region instead of got_region.json — use the .json extension for typed outputs
    inputs:
      script   ← cat ./context/region > ./outputs/got_region   (set in plan)
```

### `plan x get` — playwright run with multi-line script + artifact outputs

```
Steps:
  tests   completed
    action: playwright (948yufcflxghu)   (revision 2)
    inputs:
      script              ← const { test, expect } = require('@playwright/test');… (38 lines)   (set in plan)
      options             (unbound)
      baseURL             ← params.baseURL   (ref)
      dependencies        (unbound)
      capture_artifacts   ← true   (set in plan)
    outputs:
      test_artifacts   (artifact, 5.469MB, ready, application/gzip)
      exitCode         ← 0   (inline)
      report           (artifact, 5.073kB, ready)

Outputs:
  exitCode         ← 0                                           (inline,   from tests)
  report           (artifact, 5.073kB, ready,                    from tests)
  test_artifacts   (artifact, 5.469MB, ready, application/gzip,  from tests)
```

### `plan get` — full spec view (after section normalisation)

```
ID:           27f6u0lj6ck72
Prompt:       First use request-http action to GET http://httpbin.org/uuid. Then in a secon...
Created:      Tue, 05 May 2026 14:23:30 -03 (Less than a second ago)
Compiled From: 9tmurfk20h666

Inputs:
  region   ← us-east-1   (default)

Steps:
  get_uuid
    action: request-http (f3ppjxbzpddry)   (revision 1)
    inputs:
      method   ← GET   (set in plan)
      url      ← http://httpbin.org/uuid   (set in plan)

  check_status
    action: check (cn20hsyg8yjmb)   (revision 1)
    inputs:
      expression   ← response.statusCode == 200   (set in plan)
      name         ← Check HTTP Status Code   (set in plan)
      response     ← steps.get_uuid.outputs.capture.response   (ref)

Outputs:
  check_result   ← steps.check_status.outputs.result
```

### `plan get` — older plan (action name not yet populated)

Plans compiled before the SDK added `Name` fall back to the bare action ID:

```
Steps:
  send_request
    action: 6buzf26tv1j1q   (revision 1)
    inputs:
      method   ← GET   (set in plan)
      url      ← http://httpbin.org/uuid   (set in plan)
```

### `plan run --attach` — structured stream from creation to result

```
$ signadot plan run lmsu2671y8ufl --param sandbox=… --param baseURL=… --attach
time=19:47:11 type=created id=8y6b6w2g6ycss plan_id=lmsu2671y8ufl
time=19:47:14 type=log step=install_deps stream=stdout msg="..."
... logs streaming ...
time=19:47:47 type=output name=report step=tests kind=artifact size=5.064kB ready=true
time=19:47:47 type=output name=test_artifacts step=tests kind=artifact size=4.819MB ready=true content_type=application/gzip
time=19:47:47 type=output name=exitCode step=tests kind=inline value=0
time=19:47:47 type=result id=8y6b6w2g6ycss phase=completed
```

JSON mode (`--attach -o json`) emits the same fields as one object per line.

---

## Files touched

- `internal/command/planexec/printers.go` — value resolution, prose vias, nested step layout, action label, unified `Outputs:` section, value truncation
- `internal/command/planexec/get.go`, `cancel.go` — best-effort `GetPlan` to populate the spec
- `internal/command/plan/run.go` — pass the already-loaded plan spec through to the printer; emit `created` event in attach mode; translate `PlanOutputStatus` into `AttachEvent`
- `internal/command/plan/printers.go` — full `plan get` rebuild + action name + timeout/image, section normalisation
- `internal/command/planaction/printers.go` — drop `Requires:` and validations table that the new SDK no longer carries
- `internal/print/attach.go` — `AttachEvent` fields for `kind/size/ready/contentType/planID`, `formatText` cases for `created` and the kind-aware `output` rendering
- `go.mod` / `go.sum` — bump go-sdk to feature-plan tip with `PlanStepAction.Name`

daniel-de-vera:
@scott-cotton, latest commit (`c20239a`) tightens `plan run --attach` so the event stream stays consistent end-to-end. Two issues, both visible side-by-side in this run:

### Before

```
$ signadot plan run … --attach
Created execution 8y6b6w2g6ycss for plan lmsu2671y8ufl
time=19:47:47 type=output name=report value=<nil>
time=19:47:47 type=output name=test_artifacts value=<nil>
time=19:47:47 type=output name=exitCode value=0
time=19:47:47 type=result id=8y6b6w2g6ycss phase=completed
```

- The `Created execution …` line is a plain stderr banner; everything else is structured `time=… type=…` lines on stdout. Visual break for the user, and JSON-mode consumers piping through `jq` lost the execution ID entirely (the banner was suppressed in `-o json`).
- Artifact outputs rendered as `value=<nil>`. The full `PlanOutputStatus` was in hand at emit time (the events come from a `GetPlanExecution` snapshot at terminal phase, not from the SSE log stream) — we just weren't using it.

### After

```
$ signadot plan run … --attach
time=19:47:11 type=created id=8y6b6w2g6ycss plan_id=lmsu2671y8ufl
time=19:47:14 type=log step=install_deps stream=stdout msg="..."
... logs streaming ...
time=19:47:47 type=output name=report step=tests kind=artifact size=5.064kB ready=true
time=19:47:47 type=output name=test_artifacts step=tests kind=artifact size=4.819MB ready=true content_type=application/gzip
time=19:47:47 type=output name=exitCode step=tests kind=inline value=0
time=19:47:47 type=result id=8y6b6w2g6ycss phase=completed
```

What changed:

**`AttachEvent` extended** for output events: `kind` (`inline` | `artifact`), `size`, `ready`, `contentType`, `step`. All `omitempty` so existing JSON consumers ignoring unknown fields keep working. Format parity with the rendered `Outputs:` section in `plan x get`:
- `ready=true|false` always shown for artifacts
- `content_type=…` from `metadata.contentType` when set
- `step=<id>` for plan-level outputs that came from a step (`StepRef.StepID`)
- artifact-side `error=…` surfaced

**New `created` event** at the start of the attach stream replaces the stderr banner. Same content (execution ID + plan ID), same time/type shape as every other event. Non-attach mode still prints the banner to stderr — only `--attach` switches to structured.

### JSON mode

Consumers now get a complete event stream — no banner break, no need to filter stderr. `jq -c '.'` would give:

```json
{"time":"…","type":"created","id":"8y6b…","planID":"lmsu…"}
{"time":"…","type":"output","name":"report","step":"tests","kind":"artifact","size":5187,"ready":true}
{"time":"…","type":"output","name":"test_artifacts","step":"tests","kind":"artifact","size":5054054,"ready":true,"contentType":"application/gzip"}
{"time":"…","type":"output","name":"exitCode","step":"tests","kind":"inline","value":0}
{"time":"…","type":"result","id":"8y6b…","phase":"completed"}
```

### Files touched

- `internal/print/attach.go` — `AttachEvent` fields, `formatText` cases for `created` and the kind-aware `output` rendering
- `internal/command/plan/run.go` — `buildOutputEvent` translation from `PlanOutputStatus`, `created` event emit at start of `attachExecution`, gate the stderr banner on `!cfg.Attach`

daniel-de-vera:
@scott-cotton, latest commit (`12b6abb`) wires the new `PlanSpec.SelectionHint` field through two CLI surfaces:

### `plan tag list` — replace `PROMPT` with `SELECTION HINT`

The PROMPT column was already truncating multi-line prompts to a single line, but a typical first line is 80+ chars of cryptic context that doesn't help anyone pick between tagged plans:

**Before**

```
NAME            PLAN ID         STEPS   PROMPT                                                                             CREATED
test-headers    887r83pxutpsd   3       Send a request to svc1.test-istio:8080/ and check that the response body fiel...   29 days ago
scenario-a      02knkptjp30ww   1       Given an input namespace (a string), build the URL http://location.<namespac...   19 days ago
```

**After** (once plans get recompiled with the new SDK populating the field):

```
NAME            PLAN ID         STEPS   SELECTION HINT                                     CREATED
test-hint       797m0psk1mwwb   3       Check location service responds 200 in a sandbox   about a second ago
test-headers    887r83pxutpsd   3                                                          29 days ago
scenario-a      02knkptjp30ww   1                                                          19 days ago
```

Plans compiled before the field existed render an empty cell — graceful fallback. As plans get recompiled, the column fills in.

### `plan compile --selection-hint "..."`

The `/plans/compile` API already accepts `selectionHint` in the JSON body, but agents had to either construct the payload by hand or fall back to `plan create -f spec.yaml`. New convenience flag mirrors `--tag`:

```
$ signadot plan compile -o json -f - --selection-hint "Check location service responds 200 in a sandbox" <<'EOF'
Send a request to location.<namespace>:8081/locations (where namespace is an input)
and check the response code is 200.
EOF
```

End-to-end round-trip verified: the plan's `spec.selectionHint` carries the supplied text, prompt stays on `spec.prompt`, and `plan tag list` shows the hint when the plan is tagged.

### Files touched

- `internal/command/plantag/printers.go` — column rename + SelectionHint source + `utils.TimeAgo` cleanup
- `internal/command/plan/compile.go` — plumb `cfg.SelectionHint` into `PlanCompileInput`
- `internal/config/plan.go` — `PlanCompile.SelectionHint` field + `--selection-hint` flag
- `go.mod` / `go.sum` — bump go-sdk to feature-plan tip with the new field

scott-cotton (commented):
## Re-review

The two earlier comments were addressed cleanly: the silent filter is gone (everything is shown now, with resolved values), and `unbound` at step level surfaces instead of being suppressed. On top of that, the PR has grown into a larger rendering overhaul — full per-step layout, prose vias, value resolution via best-effort `GetPlan`, unified `Outputs:` section, attach event parity, an SDK bump, and a new `--selection-hint` flag. The output examples in the PR description look great and the core idea is right.

Build, vet, and `go test ./internal/command/planexec/...` all pass.

### Issues worth fixing before merge

**1. `go.sum` has a stale `signadot/go-sdk` entry**

```
github.com/signadot/go-sdk v0.3.8-0.20260505171357-a02f1a78bde4 h1:...
github.com/signadot/go-sdk v0.3.8-0.20260505171357-a02f1a78bde4/go.mod h1:...
github.com/signadot/go-sdk v0.3.8-0.20260505200838-3bae35b0d0d8 h1:...
github.com/signadot/go-sdk v0.3.8-0.20260505200838-3bae35b0d0d8/go.mod h1:...
```

`go.mod` only requires the `…200838` version. Running `go mod tidy` removes the `…171357` lines — looks like a leftover from the intermediate "two SDK bumps in one PR" history.

**2. Significant duplication between `internal/command/plan/printers.go` and `internal/command/planexec/printers.go`**

These helpers are now defined verbatim (or near-verbatim) in both files: `actionLabel`, `printInputLine`, `formatValue`, `truncateForDisplay`, `maxValueDisplay`, `paramsAsMap`, `stepArgsValues`. They render the same arrow/via shape and are intentionally meant to stay in sync — the comments in both copies even say so. Promoting these to a small shared package (e.g. `internal/command/planshared` or `internal/print/plan`) would prevent the two views from drifting. Not a blocker, but a meaningful follow-up.

**3. `plantag/printers.go` has a half-migrated cleanup**

The PR migrates `printTagTable` to use `utils.TimeAgo`, but `printHistoryTable` immediately below it still inlines the `time.Parse(time.RFC3339, ...) + timeago.NoMax(...)` pattern. The `time` and `timeago` imports survive only because of that one function. Migrating it in this PR would let the imports come out, and it's the same file. The PR description does mention "migrating those is left for a separate cleanup" — fair, but this one is mid-cleanup-in-progress, so it's a small finish.

### Smaller observations

- `fetchPlanSpec` in `planexec/get.go` adds a second API round-trip per `plan x get` / `plan x cancel`. The graceful-degrade behavior (return `nil`, render without resolved values) is right, but worth knowing this doubles the request count on every detail render.
- `formatOutputTag` separates the source-step suffix with `, from <stepID>`; on plan-level rollups with no other tags this reads as `(inline, from tests)`. Fine, just slightly awkward — one example in the PR description even has it landing inside what visually looks like multi-column padding (`(artifact, 5.073kB, ready,                    from tests)`). Worth eyeballing in a real terminal to confirm it doesn't look like a formatting bug.
- `printPlanInputs` takes `ex *models.PlanExecution` but only uses `ex.Status.Inputs` and `ex.Spec.Params` (via `execParams`). Could narrow the parameter, but not a real problem.
- No tests for the new printer logic. Given how much branching there is (inline vs artifact, with/without StepRef, multi-line truncation, action label fallbacks, all four resolved-via prose forms), a few golden-file tests would buy a lot of regression coverage cheaply. Out of scope for this PR if you'd rather follow up.

### Overall

Good iteration on the prior feedback, and the merged output shape is clearly better than what was there. The `go.sum` stale entry is the thing I'd most want addressed in this PR; the duplication note is more of a design heads-up that I'd push as a follow-up.

daniel-de-vera:
@scott-cotton, pushed https://github.com/signadot/cli/pull/326/changes/5aa7baf8fb7dc9bdf3495415a5f52bc9292bd886. Review items addressed: 

1. go.sum stale entries — go mod tidy removed the stale …171357 lines.
2. Duplication — new internal/command/planshared/render.go package owns PrintInputLine, FormatValue (+ private truncateForDisplay/MaxValueDisplay), FormatImage, ActionLabel, ParamsAsMap, StepArgsValues.
 plan, planexec, planaction all import it; ~265 lines of duplication deleted. 
3. plantag half-migrated cleanup — printHistoryTable now uses utils.TimeAgo; time and timeago imports dropped from the file.
4. Smaller observation: printPlanInputs parameter — narrowed to (inputs, suppliedParams, planSpec); execParams helper gone. 
5. Smaller observation: formatOutputTag formatting — verified live output is compact; the misleading "padding" in the PR comment was just my manual whitespace in a rendered example, not a code bug.

scott-cotton (approved):
LGTM

PR #328, opened 2026-04-30
plan run --attach: resolve drill-in outputs and render values as JSON

daniel-de-vera:
## Summary

Two related fixes to `plan run --attach` that improve the streamed output UX. Both are in the same code path — the end-of-execution output-event emission in \`internal/command/plan/run.go\`.

### 1. Drill-in (and artifact-backed) outputs streamed as \`value=<nil>\`

Plan-level outputs that drill into a JSON sub-field (e.g. \`steps.send.outputs.capture.response.statusCode\`) and outputs whose whole value lives in S3 leave \`PlanOutputStatus.Value\` nil. The runner deliberately records the ref + drill path without materializing the resolved value — the apiserver applies the drill at serve time inside \`GetPlanExecutionOutput\`. From the type comment:

> The drill is applied at download time by the apiserver — the runner records it but does not materialize the drilled value, so the source step's existing artifact (or inline Value) is reused without any extra upload.

The attach flow was reading \`o.Value\` directly, so every drill-in surfaced as \`value=<nil>\` while \`plan x get-output <exec> <name>\` returned the right value afterward. Same path for artifact-backed whole outputs (Value nil because the bytes live in S3).

**Fix:** at end of stream, for any output with nil Value, fetch the resolved bytes via \`GetPlanExecutionOutput\` and parse as JSON (or fall through to the raw string for non-JSON outputs). Best-effort — per-output fetch failures are logged to stderr but don't abort the stream, so worst case the behavior degrades to today's nil-emission.

### 2. Object-typed values rendered as Go's \`map[k:v]\` instead of JSON

With (1) fixed, any output that resolves to a map/slice was rendered via \`fmt.Sprint\`, producing \`map[check:frontend / returns 200]\` instead of structured JSON. Inconsistent with how a user reading the stream expects to see structured values.

**Fix:** new \`formatAttachValue\` helper. Strings stay unquoted (the common case for plan outputs); nil renders as \`<nil>\` matching prior behavior; everything else (numbers, bools, maps, slices) is JSON-marshaled. So \`{"check":"frontend / returns 200"}\` instead of \`map[check:...]\`, and \`200\` instead of \`200\` (already worked, but now via the same code path).

## Test plan

- [x] \`go build ./...\` clean.
- [x] \`go vet ./...\` clean.
- [x] \`gofmt -l\` clean.
- [x] \`go test ./internal/print/\` — new \`TestFormatAttachValue\` covers nil, string, empty string, number, bool, map, slice, nested object (9 cases, all pass).
- [ ] Manual repro against dev: run a plan with a drill-in output and an object-typed output via \`--attach\`; confirm the streamed \`value=\` for both (a) renders the resolved value rather than \`<nil>\`, and (b) renders the object as \`{"check":"..."}\` rather than \`map[check:...]\`. The plan from sandboxes investigation (\`steps.send_request.outputs.capture.response.statusCode\` + \`steps.assert_200.outputs.result\`) reproduces both fixes in one execution.

## Notes

- The resolution path (1) does N extra API calls at end-of-stream, where N = number of plan-level outputs whose Value was nil. Tiny in practice; bounded by the existing 1 MB drill-source cap.
- The \`--output-dir\` export path already calls \`GetPlanExecutionOutput\` per output; this generalizes the same pattern to attach.

### DIFF
--- a/internal/command/plan/run.go
+++ b/internal/command/plan/run.go
@@ -1,6 +1,7 @@
 package plan
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -20,8 +21,8 @@ import (
 	sdkclient "github.com/signadot/go-sdk/client"
 	planlogs "github.com/signadot/go-sdk/client/plan_execution_logs"
 	planexecs "github.com/signadot/go-sdk/client/plan_executions"
-	sdkplans "github.com/signadot/go-sdk/client/plans"
 	plantags "github.com/signadot/go-sdk/client/plan_tags"
+	sdkplans "github.com/signadot/go-sdk/client/plans"
 	"github.com/signadot/go-sdk/models"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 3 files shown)


daniel-de-vera:
Closing this in favor of https://github.com/signadot/cli/pull/326.

PR #330, opened 2026-05-07
trafficwatch: drop safeCloser, use TrafficWatch.JustClose()

scott-cotton:
## Summary
Fixes a `panic: close of closed channel` from `libconnect/common/trafficwatch.WatchTrafficClient.Watch.func1.1` reproducible under e2e + `signadot traffic record` against staging.

`setupTW` (in `internal/trafficwatch/get.go`) wrapped `tw.Close` in a local `safeCloser` whose own `sync.Once` was unrelated to the lib's `closeOnce`. The wrapper's "is it closed yet?" probe (`select { case <-ch: default: close(ch) }`) isn't atomic with the close, so two distinct Onces could each reach `close(ch)`:

1. lib goroutine enters `closeCOnce.Do(...)`, grabs the mutex, starts running `f`, about to `close(closeC)`.
2. `setupTW`'s goroutine enters its own `once.Do(...)`; the `<-ch` non-blocking probe sees the channel still open (the lib hasn't closed it yet — only acquired the mutex), falls through to `default`, and calls `close(ch)`.
3. lib goroutine's `f` runs `close(closeC)` → panic.

The lib's `TrafficWatch.JustClose()` is exactly the right primitive: same `closeOnce` as the lib's five internal `close(closeC)` call sites, so they all coordinate. The docstring on `TrafficWatch.Close` in libconnect explicitly warns against `close(tw.Close)` directly.

This PR drops `safeCloser` and its dedicated test; `setupTW` now calls `tw.JustClose()` on signal / timeout.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`


daniel-de-vera (approved):
LGTM

PR #332, opened 2026-05-13
Cli versioned resource plugins

scott-cotton:
> Depends on:
> - https://github.com/signadot/signadot/pull/7007 (server-side symmetric-name shape)
> - https://github.com/signadot/go-sdk/pull/80 (regenerated client from the above)

This PR addresses https://github.com/signadot/community/issues/116

## What changes for users

Plugin references take the form `name[@<semver>]` everywhere a user types or sees a plugin reference:

```
signadot rp apply -f my-plugin.yaml       # version comes from the YAML name: field
signadot rp list                          # NAME column renders foo@1.2.0 (or bare foo at default)
signadot rp versions my-plugin            # new: every published version
signadot rp get my-plugin                 # latest
signadot rp get my-plugin@1.2.0           # specific version
signadot rp delete my-plugin@1.2.0
```

In YAML, the version travels with `name:` as an `@<semver>` suffix — there is no separate `version:` field. A bare name publishes the plugin at the default version (`0.0.0`):

```yaml
name: my-plugin@1.2.0
spec: { ... }
```

A stray top-level `version:` field is silently ignored, like any other unknown top-level field.

## What `get` / `list` look like

For plugins published at the default version, `get -o yaml` returns the bare `name: my-plugin` (no `@0.0.0` suffix), matching the pre-versioning CLI surface byte-for-byte. For any concretely-versioned plugin, the response carries the version in the `name` field as a single token. `get -o yaml` round-trips back into `apply -f` without renaming.

## Server contract

- Identity on the wire is a single `name` field; the version is the optional `@semver` suffix.
- Versions are immutable: re-applying an existing `(name, version)` — including the implicit default `0.0.0` — returns 409. The CLI maps that to `resource plugin %q version %q already exists; versions are immutable — bump the version field to publish a new revision`, and uses `default version (0.0.0)` when the user never set one.
- `get` and `delete` accept `?version=<v>` or omit/`?version=latest` for the highest semver.
- `ListResourcePlugins` returns latest-only; per-plugin version listing is the new `versions` subcommand.

## Test plan

- [x] `go test ./...` passes locally.
- [x] Loader unit tests cover bare name, `@version` suffix, and the now-ignored top-level `version:` field.
- [ ] `--help` for `resourceplugin`, `get`, `versions` renders the new `NAME[@VERSION]` usage.
- [ ] e2e: apply a plugin with bare name, GET it back, see `name: my-plugin`.
- [ ] e2e: apply a plugin with `name: foo@1.2.0`, list, get by version, list versions, delete by version.
- [ ] e2e: apply same plugin twice with same version → see the 409 message.

## Before merging

- Re-pin go-sdk to the merged tag from go-sdk#80 (currently on the `versioned-resource-plugins` branch tip).


scott-cotton:
The latest push (3 commits, `23bc855..7577984`) adapts this PR to the symmetric-name wire shape introduced in signadot/signadot#7007. Summary of what changed from the previous state of the branch:

- **`name@version` is the wire form, end to end.** The previous shape carried `name` and `version` as two body fields on PUT and as two response fields on GET. Now the body and the response both use a single `name` field of the form `bareName[@semver]`. `get -o yaml` pastes straight back into `apply -f`.
- **Top-level YAML `version:` is dropped.** The earlier review feedback had us accept either `name: foo@1.2.0` *or* `name: foo` + `version: 1.2.0`. Once the wire identity collapsed to a single token, supporting two YAML forms for the same thing only widened the input space beyond what `get -o yaml` produces. The loader now reads `name:` as the wire form directly; a stray `version:` key is silently ignored, like any other unknown top-level field. Users who hand-wrote the two-field form should fold the version into the `name` suffix.
- **Default-version plugins display bare.** A plugin published at `0.0.0` (no `@semver` in `apply`) is shown as `name: my-plugin` everywhere — the CLI's `apply` success message, `get`/`list`/`versions` output, and the JSON body the server returns. Concrete versions still render `name: my-plugin@1.2.0`. This preserves byte-for-byte compatibility with the pre-versioning CLI surface for the typical "I don't care about versioning" user.

The CLI commits in this push correspond 1:1 to the three layers of change:

1. `23bc855 resourceplugin: adapt to combined name wire field` — go-sdk bump; split `req.Name` for the URL-path arg in `apply`/`delete`; use `rp.Name` directly in printers (it's already joined on the wire).
2. `2577541 resourceplugin: drop top-level YAML version field` — remove `optionalStringField` and the XOR check from `subst.go`; trim related test cases; drop the `version` key from `utils.extractName`'s keep-list for `delete -f`.
3. `7577984 resourceplugin: render default-version plugins as bare name` — go-sdk re-bump for the bare-default server change; drop `effectiveNameRef` in favor of `formatNameRef` (which already produces bare when version is empty).

Net: this PR no longer adds any local string-splitting "cleverness" to paper over a server asymmetry — the server now exposes the same shape the CLI always wanted.

scott-cotton:
> ```
> NAME        CREATED                       STATUS
> foo@2.0.0   2026-05-19T14:50:09.656114Z   0 resources
> foo@1.5.0   2026-05-19T14:53:39.50125Z    0 resources
> foo@1.1.0   2026-05-19T14:50:09.055472Z   0 resources
> foo@1.0.0   2026-05-19T14:50:07.937688Z   0 resources
> ```
> 
> IMO this is verbose, we are repeating the same word over an over (`foo`)

Yes but IMO it should be verbose because this is tabular output and hence both expected and easier to automate, which in my mind outweighs the repetition.  wdyt in light of this?


> 
> Also the `CREATED` time doesn't seem to be aligned with others, for example `sb list`

This PR did not change the time, it was pre-existing problem.  But, I agree it would be better to be aligned, so I've added that in https://github.com/signadot/cli/pull/332/commits/5577ebaeee16e76f896b260a3a7184671b509e09.

daniel-de-vera (approved):
LGTM

PR #333, opened 2026-05-18
devbox: self-heal stale cached ID on 404

scott-cotton:
## Summary
- `~/.signadot/.devbox-id` becomes stale when the server GCs an idle devbox (30-day TTL in `sandboxes/internal/control/devbox/constants.go`). Every subsequent CLI call then warns or fails with a 404.
- `devbox.GetID` now treats a 404 from `GetDevbox` or `ClaimDevbox` the same as a missing cache file: re-register and overwrite the file. Per-user devboxes recover transparently on the next `signadot local connect`.
- `devbox list` no longer prints the scary `404 Not Found: no such devbox …` warning when the cached ID is gone — it just omits the "current" row from the table. A `--debug` hint identifies the cleared id for troubleshooting.
- The explicit `signadot local connect --devbox <id>` path is **unchanged**: a user-supplied ID still errors hard on 404 rather than silently recreating.

Fixes signadot/community#136

### Behavior trade-off
There's a competing use case where a platform team pre-registers a userless devbox on a Remote Dev Environment so multiple users share that RDE's local-connect session slots. Under that pattern, a loud failure when the pre-provisioned devbox vanishes would arguably be preferable to silently recreating one. We accept the trade-off because RDE devboxes very rarely sit 30 days idle without use, and the per-user case is hit constantly. If this pattern grows, the right follow-up is a config knob or devbox-metadata marker to opt into strict (no-auto-recreate) behavior.

### Implementation notes
- New `devbox.IsNotFound(err)` predicate unwraps to `*transport.APIError` (not `*runtime.APIError`) — `transport.FixAPIErrors` middleware in the SDK wraps every 4xx/5xx before the swagger reader sees it.
- New unexported `registerAndPersist` helper centralizes the "register and write cache file" step, used both by the original file-not-exist path and the new 404-recovery sites.

## Test plan
Verified end-to-end against a local minikube + signadot install with apiserver backed by host MySQL:

- [x] Register a fresh devbox via `signadot devbox register`; confirm row in MySQL and matching `~/.signadot/.devbox-id`.
- [x] `DELETE FROM devboxes WHERE public_id=…;` to simulate the GC step.
- [x] `signadot devbox list` (no `--debug`): silent, no warning, table renders without the stale row.
- [x] `signadot --debug devbox list`: prints `debug: cached devbox <id> is no longer registered`.
- [x] `signadot local connect --unprivileged --wait=none --cluster=minikube`: completes silently; cache file rewritten with a fresh valid id; session claimed.
- [x] `go build ./...` and `go vet ./internal/devbox/... ./internal/command/devbox/... ./internal/command/local/...` clean.

### DIFF
--- a/internal/command/devbox/list.go
+++ b/internal/command/devbox/list.go
@@ -72,7 +72,17 @@ func list(cfg *config.DevboxList, out io.Writer, errOut io.Writer) error {
 		if !currentDevboxInList {
 			ddb, err := get(ctx, cfg, currentDevboxID)
 			if err != nil {
-				fmt.Fprintf(errOut, "Warning: Could not fetch devbox %s from ~/.signadot/.devbox-id: %v\n", currentDevboxID, err)
+				switch {
+				case devboxpkg.IsNotFound(err):
+					// Cached devbox no longer exists server-side (e.g. 30-day
+					// idle TTL). GetID self-heals on the next operation that
+					// needs it; here we just treat it as absent.
+					if cfg.Debug {
+						fmt.Fprintf(errOut, "debug: cached devbox %s is no longer registered\n", currentDevboxID)
+					}
+				default:
+					fmt.Fprintf(errOut, "Warning: Could not fetch devbox %s from ~/.signadot/.devbox-id: %v\n", currentDevboxID, err)
+				}
 				// set current to unknown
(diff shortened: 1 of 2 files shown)


foxish (approved):
LGTM

PR #342, opened 2026-06-09
local connect: add opt-in --local-dns for resolving cluster names via a local DNS resolver

scott-cotton:
## What
Adds opt-in local-DNS resolution to `signadot local connect`. When enabled, the CLI resolves Kubernetes service names through a **local DNS resolver** wired into your OS resolver, instead of the default mechanism that edits `/etc/hosts`.

Two ways to opt in:

- **Per connection (encouraged)** — set `resolver: LocalDNS` on the connection in your config. `resolver: EtcHosts` (the default when the field is omitted) keeps the `/etc/hosts` behavior:
  ```yaml
  local:
    connections:
      - cluster: <cluster>
        resolver: LocalDNS
  ```
- **Per invocation** — the `--local-dns` flag overrides the config field for a single connect. `--local-dns=false` forces `/etc/hosts` even when the config selects `LocalDNS`:
  ```
  signadot local connect --local-dns --cluster <cluster>
  ```

Default behavior is unchanged — omitted config field and no flag = current `/etc/hosts` behavior, byte-for-byte. This is additive and experimental.

## Why
The default path enumerates a fixed set of hostnames per service and rewrites the shared `/etc/hosts` file on every cluster change. Local DNS instead runs a real resolver that is authoritative for the cluster zones: it answers cluster names directly, forwards everything else to your existing upstream resolvers, and restores your DNS configuration on disconnect. If a session exits uncleanly, DNS can be left pointed at the stopped resolver; the **next `signadot local connect` repairs it on startup regardless of mode** — recovery is no longer gated on `--local-dns`, so a plain reconnect (the natural move when your DNS just broke) restores working resolution. Net effect for users: cluster name resolution that behaves more like in-cluster DNS, without touching `/etc/hosts`.

## What resolves
With local DNS active, all the usual service name forms resolve on **both macOS and Linux**:

- `my-svc.my-ns.svc.cluster.local` (FQDN)
- `my-svc.my-ns.svc`
- `my-svc.my-ns`

Non-cluster names (e.g. `github.com`) continue to resolve normally — they're forwarded to the resolvers you were already using.

## `local hosts`
`signadot local hosts` lists the cluster hostnames currently resolvable from your machine and the address each maps to. It reads the shared address map, so it works in **both** resolution modes (`/etc/hosts` and local DNS):

```
signadot local hosts            # aligned NAME / IP table
signadot local hosts -o json    # [{ "name": ..., "ip": ... }, ...]
signadot local hosts -o yaml
```

Each host reports a single virtual address (IPv4-preferred); the addresses are the synthetic ones the tunnel intercepts, deterministic and name-sorted.

## Requires root
Local DNS configures the system DNS resolver (and on Linux binds `127.0.0.54:53`), so it needs the same root privileges `local connect` already escalates for. The privilege-rationale prompt now mentions DNS-resolver configuration when active. It is **not** supported with `--unprivileged` — combining them fails fast with a clear error rather than silently ignoring the setting (there's no root-manager in that mode to run the resolver). This applies however local DNS was selected: the same conflict check fires when the mode comes from the `resolver:` config field, and the error names the source.

## `local status`
When local DNS is active, `signadot local status` shows the resolver in place of the `/etc/hosts` line — resolvable record count, the resolver's bind address, and a prominent warning if a system DNS manager (e.g. systemd-resolved owning `/etc/resolv.conf`) might revert the configuration. The `/etc/hosts` section is omitted in this mode rather than reported as a failure. Reflected in both the table and JSON/`--details` output.

## Compatibility & scope
- **Opt-in:** omitted config field and unset flag = current `/etc/hosts` behavior, byte-for-byte.
- **Platforms:** macOS and Linux. (Other OSes reject resolver management.)
- **Intentional difference vs `/etc/hosts`:** the resolver mirrors what a real cluster serves, so it will **not** resolve non-standard truncations the old `/etc/hosts` path happened to emit (e.g. `my-svc.my-ns.svc.cluster`) — those never resolve in-cluster either. Surfacing this as a deliberate behavior change since it's opt-in.
- **IPv6 — groundwork only:** the underlying libconnect work includes first steps toward IPv6 (a dual-stack address-allocation model and an AAAA-capable resolver), but it is **not yet employed** by either the cluster or the CLI/client side — AAAA queries return NODATA for now. `local hosts` accordingly reports one address per host; the resolver's `[]ip` backing (for future A/AAAA answers) is unaffected.
- **Namespace-qualified bare names:** `my-svc` → current namespace, not yet.

## Try it
```
signadot local connect --local-dns --cluster <cluster>
curl my-svc.my-ns:<port>/...        # short name
curl my-svc.my-ns.svc.cluster.local:<port>/...
signadot local hosts                # resolvable cluster names and their addresses
signadot local status               # shows the local DNS resolver
signadot local disconnect           # restores your original DNS config
```

## ⚠️ Merge order
This branch pins `libconnect` to an **unmerged** branch pseudo-version (`feat/localdns-1f`, signadot/libconnect#163) that carries the resolver, the standalone crash-recovery entry point, and the per-connection `resolver` config type. Do not merge until that lands; then repoint `go.mod` to the merged libconnect commit/tag. Kept as a draft for this reason.


internal/locald/rootmanager/tp_monitor.go
```diff
@@ -160,15 +160,17 @@ func (mon *tpMonitor) checkTunnelProxyAccess(ctx context.Context) (success bool,
...
 	// Restart localnet
 	mon.root.stopLocalnetService()
 	mon.root.runLocalnetService(ctx, mon.tpLocalAddr, mon.ipMap)
 
-	// Restart etc hosts
-	mon.root.stopEtcHostsService()
-	mon.root.runEtcHostsService(ctx, mon.tpLocalAddr, mon.ipMap)
+	// Restart name resolution (localdns or etchosts)
+	mon.root.stopNameService()
+	if err := mon.root.runNameService(ctx, mon.tpLocalAddr, mon.ipMap); err != nil {
```
foxish, line 171: **Blocking: restart loop in `--local-dns` mode.** `checkRootServer` (L206) still gates health on the /etc/hosts service: `if r.etcHostsSVC == nil || !r.etcHostsSVC.Status().Healthy`. With `--local-dns`, `etcHostsSVC` is always nil (`localDNSSVC` is set instead), and this monitor runs for exactly the PortForward and ControlPlaneProxy link types. So once the startup grace period (`connectTimeout`, default 10s) expires, every check returns `restart=true`, and localnet + localdns get torn down and restarted every ~10s indefinitely, uninstalling and reinstalling the OS resolver config each cycle. It also means localdns health is never actually monitored.

`checkRootServer` needs to branch on `ciConfig.EnableLocalDNS` and check `localDNSSVC` health instead, ideally via the mutex-guarded accessors (the raw field reads here are a pre-existing race that now extends to the new fields).

internal/locald/rootmanager/tp_monitor.go.head
```diff
@@ -0,0 +1,218 @@
+package rootmanager
```
foxish: **Blocking (trivial):** this looks like a leftover merge-conflict artifact, an old copy of `tp_monitor.go` (it still has the pre-`connectTimeout` `maxStartingTime` logic). The extension keeps it out of the build, but it should be deleted.

internal/locald/rootmanager/root_manager.go
```diff
@@ -77,9 +82,12 @@ func (m *rootManager) Run(ctx context.Context) error {
 		// starting/restarting the localnet and etchost services
 		m.tpMonitor = NewTunnelProxyMonitor(ctx, m, ipMap)
 	default:
-		// Start localnet and etchost services
+		// Start localnet and the name-resolution service (localdns or etchosts)
 		m.runLocalnetService(ctx, m.ciConfig.ConnectionConfig.ProxyAddress, ipMap)
-		m.runEtcHostsService(ctx, m.ciConfig.ConnectionConfig.ProxyAddress, ipMap)
+		if err := m.runNameService(ctx, m.ciConfig.ConnectionConfig.ProxyAddress, ipMap); err != nil {
+			_ = m.stopLocalnetService()
+			return fmt.Errorf("error starting name resolution: %w", err)
```
foxish, line 102: **Blocking: orphaned sandboxmanager on this error path.** This early return skips the cleanup block at the end of `Run()`, notably `m.sbmMonitor.stop()`. The sandbox manager was already spawned above (`go m.sbmMonitor.run()`) as a separately daemonized child via `sudo -u <user>`, and `stop()` is the only thing that kills it. So a localdns startup failure here (realistic case: something else already bound 127.0.0.54:53 on Linux) exits the root manager but leaves an orphaned sandboxmanager running with its pidfile held. Before this PR this branch had no error path, so this is new. Suggest routing the error through the same cleanup used by the normal shutdown path below.

internal/config/local.go
```diff
@@ -115,6 +116,8 @@ func (c *LocalConnect) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&c.Cluster, "cluster", "", "specify cluster connection config")
 
 	cmd.Flags().BoolVar(&c.Unprivileged, "unprivileged", false, "run without root privileges")
+	cmd.Flags().BoolVar(&c.LocalDNS, "local-dns", false,
```
foxish, line 119: We'd like the config file to be the encouraged way to opt in, with this flag as a per-invocation override: forgetting the flag on the next connect is the natural failure mode, and a `--local-dns` session followed by a plain reconnect is also exactly the case where crash recovery doesn't run (see review body).

Concrete proposal: a per-connection field in libconnect's `ConnectionConfig`, e.g.

```yaml
local:
  connections:
    - cluster: demo
      resolver: LocalDNS   # or EtcHosts; omitted = EtcHosts
```

Enum rather than bool (leaves room for `Auto` later), PascalCase values to match `type:`, zero value `EtcHosts` so existing configs are untouched. `runConnect` already has the `ConnectionConfig` in hand before building the ciConfig, so the CLI plumbing is small; the `--unprivileged` conflict check should then also fire when the mode comes from config, with an error naming the config field. Fine as a fast-follow alongside libconnect#163, but worth agreeing on the field now so docs can lead with it.

internal/command/local/hosts.go
```diff
@@ -0,0 +1,98 @@
...
+	return cmd
+}
+
+// printableHost is the JSON/YAML shape of a single resolvable host. Each host
+// carries exactly one address (see pickIP), which matches how the cluster
+// assigns names: a single-stack cluster hands out one family, and even when
+// both are present a resolver answers a name with one address at a time.
+type printableHost struct {
+	Name string `json:"name"`
+	IP   string `json:"ip"`
```
foxish, line 34: Schema forward-compat: this bakes a single `ip` into the `-o json`/`-o yaml` output, while the proto already carries `repeated ips` and this PR lays IPv6 dual-stack groundwork. When AAAA support lands, either `ip` gets a breaking rename to `ips` for script consumers, or it stays and silently under-reports dual-stack hosts. Emitting `"ips": [...]` from day one avoids the breaking change; the human-readable view can keep showing one address.

internal/command/local/hosts.go
```diff
@@ -0,0 +1,98 @@
...
+		if fallback == "" {
+			fallback = ip
+		}
+	}
+	return fallback
+}
+
+// printHosts writes one "<fqdn> <ip>" line per host. The entries arrive already
+// sorted by name from the root controller (see rootServer.GetHosts).
+func printHosts(out io.Writer, hosts []printableHost) error {
```
foxish, line 71: Default output should use `internal/sdtab` like every other list command (`cluster`, `routegroup`, `resourceplugin`, ...), so humans get an aligned NAME/IP table with a header, and scripts use `-o json`. The bare `"%s %s"` lines are also not hosts-file format (that's `<ip> <name>`, reversed), so there's no copy-paste argument for keeping them. Following the pattern from `cluster/printers.go`:

```go
type hostRow struct {
	Name string `sdtab:"NAME"`
	IP   string `sdtab:"IP"`
}

func printHosts(out io.Writer, hosts []printableHost) error {
	t := sdtab.New[hostRow](out)
	t.AddHeader()
	for _, h := range hosts {
		t.AddRow(hostRow{Name: h.Name, IP: h.IP})
	}
	return t.Flush()
}
```

internal/config/locald.go
```diff
@@ -76,6 +76,10 @@ type ConnectInvocationConfig struct {
 	Env              []string                     `json:"env"`
 	VirtualIPNet     string                       `json:"virtualIPNet"`
 	ConnectionConfig *connectcfg.ConnectionConfig `json:"connectionConfig"`
+
+	// EnableLocalDNS runs the libconnect localdns resolver (managing the OS
+	// resolver config) in place of the /etc/hosts mechanism for cluster names.
+	EnableLocalDNS bool `json:"enableLocalDNS,omitempty"`
```
foxish: gofmt: this insertion breaks the struct-tag alignment for the rest of the field block; `gofmt -l` flags this file on the branch (it's clean on main).

internal/command/local/printers.go
```diff
@@ -219,6 +221,46 @@ func getRawHosts(cfg *config.LocalStatus, ciConfig *config.ConnectInvocationConf
 	return result
 }
 
+func getRawLocalDNS(cfg *config.LocalStatus, ciConfig *config.ConnectInvocationConfig,
```
foxish, line 231: The inverse gating is missing on `getRawHosts`: in `--local-dns` mode the root manager reports `Hosts == nil`, so the non-details JSON/YAML status output still emits `"hosts": {"healthy": false, "numHosts": 0}`, which reads as a failure even though /etc/hosts management is intentionally not running. Suggest `getRawHosts` return nil when `ciConfig.EnableLocalDNS`, mirroring how this function returns nil when local DNS is disabled. (The human-readable view is already handled by the `printSuccess` branch.)

internal/locald/sandboxmanager/sdk.go
```diff
@@ -38,6 +38,23 @@ func GetStatus() (*sbmapi.StatusResponse, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	defer grpcConn.Close()
+
+	// get the hosts
+	sbManagerClient := sbmapi.NewSandboxManagerAPIClient(grpcConn)
+	resp, err := sbManagerClient.GetHosts(context.Background(), &sbmapi.GetHostsRequest{})
+	if err != nil {
+		return nil, processGRPCError("unable to get hosts from sandboxmanager", err)
```
foxish, line 61: Tested against a running daemon from a previous CLI release: this surfaces as `tunnel-api unsupported method: consider upgrading the operator`, because `processGRPCError` maps `codes.Unimplemented` to operator guidance. For `GetHosts` the call is CLI-to-local-daemon, and `Unimplemented` means the running locald predates this CLI, which will happen to everyone who upgrades the CLI while connected. Worth catching `Unimplemented` here with a targeted message, e.g. "the running local daemon predates 'local hosts'; run signadot local disconnect && signadot local connect".

foxish (commented):
Reviewed and tested this locally: all packages build, `go vet` is clean, the full test suite passes, and I smoke-tested the new CLI surface (`--local-dns` + `--unprivileged` fails fast as intended, help text, `local hosts` output paths). Nice work overall: the runNameService/stopNameService split is clean, teardown ordering (resolver restored before the tunnel drops) is well thought out, and `localdns.New` failing synchronously on bind conflicts while refreshing records asynchronously gives the right failure semantics.

Inline comments cover the code-level items. Three of them I'd consider blocking: the tpMonitor restart loop in `--local-dns` mode, the stray `tp_monitor.go.head` file, and the orphaned sandboxmanager on the new startup error path.

Cross-cutting items that don't anchor to a single line:

**1. Crash recovery only runs if the next connect also passes `--local-dns`.** The repair logic lives in libconnect's `localdns` installer and only executes inside `localdns.New`, which the CLI only calls when `EnableLocalDNS` is set. On Linux, a `--local-dns` session that dies uncleanly leaves `/etc/resolv.conf` pointing at a dead 127.0.0.54:53, i.e. all DNS on the machine is broken; if the user then runs a plain `signadot local connect` (the default, and the likely move for someone whose DNS just broke), recovery never runs and system DNS stays broken. That contradicts the description's claim that "the next `signadot local connect` detects the leftover configuration and repairs it on startup". Suggest exporting a standalone repair entry point from libconnect (feedback for signadot/libconnect#163 while it's still open) and calling it unconditionally at root-manager startup in both modes; at minimum the PR description should be corrected.

**2. `signadot local hosts` is a new user-facing command but isn't mentioned in the PR description.** It's independently useful (works in both resolution modes since it reads the shared ipmap); it should be described here and get docs.

**3. No tests accompany the change.** `pickIP`, `checkLocalDNSStatus`, and the status-printer gating are pure functions that are cheap to unit test, and a mode-matrix test on the tpMonitor health check would have caught the restart loop.

Also: we'd like the config file, not the flag, to be the encouraged way to opt in; see the inline comment on the flag definition for a concrete field proposal (`resolver: EtcHosts | LocalDNS` per connection).

internal/locald/rootmanager/tp_monitor.go
foxish: Confirmed empirically: built this branch, connected to a real cluster with `--local-dns` over a PortForward link. The daemon restarts localnet + localdns every ~10s, continuously (observed for over an hour), binding a new ephemeral port each cycle and rewriting all 12 `/etc/resolver` files + HUPing mDNSResponder each time:

```
15:50:41 INFO msg="restarting localnet and name-resolution services"
15:50:41 INFO msg="localdns listening" addr=127.0.0.1:55286
15:50:51 INFO msg="restarting localnet and name-resolution services"
15:50:51 INFO msg="localdns listening" addr=127.0.0.1:63746
15:51:01 INFO msg="restarting localnet and name-resolution services"
15:51:02 INFO msg="localdns listening" addr=127.0.0.1:53087
...
```

Easy to miss because `local status` reports healthy between restarts and resolution genuinely works (FQDN/short forms resolve, end-to-end curl through localnet returns 200).

internal/command/local/hosts.go
scott-cotton: My reasoning for the standard output was that this command is different because it is likely to get many many more outputs than another list using sdtab.  So, the headers basically are likely to be invisible, off screen in any case -- why force the acrobatics and extra machinery of json or header filtering on consumers?

More generally, removing the headers is a basic UX concern regarding standard unix pipelining.

But, since it seems this reasoning only subtly justifies deviating from the unusable-in-pipelines UX elsewhere, I'll add the sdtab.

internal/command/local/hosts.go
scott-cotton: I will change the proto to be singleton ip.  This doesn't affect ipv6/dual-stack forward compatibility and is simpler, following the table below:

| ipv4 | dual-stack | ipv6 |
|-|-|-|
| use ipv4 | use ipv4 (both virtual addrs reach the service; v4 by preference) | use ipv6 |

the ipMap representation retains `[]ip` as it is the backing for the resolver which needs to answer AAAA and A record requests.

internal/locald/sandboxmanager/sdk.go
scott-cotton: oops.  good catch!

internal/config/local.go
joehorsnell: This is a good idea @foxish.

One question here @scott-cotton - do we still check at runtime, or record at the actual point of connecting, which resolver method was used, rather than relying on the config file/command-line switches to tell us?

I'm just thinking of the case where the value in the config file changes while the CLI is connected, eg someone connects using the default `/etc/hosts` method, then while connected modifies the config file to use the `LocalDNS` setting.

scott-cotton: The CLI is designed to have race-free local connect config, it records the config at invocation time and sends that to the locald process, thus making it impossible to unexpectedly change the config or be unaware of what the config is due to it changing while running.

this is actually necessary because the config includes, for example, the auth (and apiserver) and the possibility of having parse errors, of cli invocation overriding the config, etc.   All these things make the expected flow quite complex for a live config file.

So, if someone changes anything in the config, it has no effect and they should restart, which IMO is by far the least friction approach.

foxish (approved):
Re-reviewed the latest state (`f2ebe46`, libconnect pinned at `c142183` which matches the current `feat/localdns-1f` head) and ran it live against the demo cluster on macOS: PortForward link, `--local-dns`, 15+ minute session.

Everything from the previous round is verifiably fixed, in code and live:

- No restart loop: exactly one "restarting localnet and name-resolution services" at startup (the expected initial start when the port-forward address first materializes), then stable for the whole session on one bind address, resolver-file mtimes frozen at daemon start.
- All three name forms resolve through the real macOS resolver path (`dscacheutil`, i.e. getaddrinfo): FQDN, `<svc>.<ns>.svc`, `<svc>.<ns>`. External names unaffected.
- Correct DNS semantics probed directly: A answers authoritatively; AAAA is authoritative NODATA (the documented IPv6 groundwork behavior); a miss under `cluster.local` is an authoritative NXDOMAIN; a miss under a borrowed namespace zone is forwarded upstream (verified against captured Tailscale MagicDNS upstreams), so the short-name zones don't shadow real domains. TCP works.
- Hygiene held: `/etc/hosts` has zero signadot entries in this mode, all managed files carry the sentinel, and a pre-existing foreign `/etc/resolver/test` file on my machine was correctly left untouched. `scutil --dns` shows all scoped resolvers registered.
- `local hosts` (table/JSON), `local status` (table/JSON/`--details`) all gate correctly per mode; clean disconnect left nothing behind.

**Approving, modulo the two bugs below being addressed.** Both live in the pinned libconnect branch rather than this repo, so practically: fix in signadot/libconnect#163 and bump the pin here, which the merge-order note requires before undrafting anyway.

### 1. Linux: a stale sentinel resolv.conf without its backup produces a self-forwarding loop

The chain: a crashed `--local-dns` session leaves `/etc/resolv.conf` signadot-managed (sentinel + `nameserver 127.0.0.54`), and `/var/lib/signadot/resolv.conf.bak` is missing (wiped dir, failed backup write). On the next connect:

1. `localdns.Recover` no-ops, because Linux recovery is keyed on the backup file existing (`installer_linux.go recover()`).
2. `localdns.New` binds 127.0.0.54 (free, the old process is dead).
3. `captureUpstreams` reads the stale resolv.conf and captures **our own address** as the upstream. Every non-cluster query now self-forwards: each forwarded query arrives as a new inbound query, so this is a query loop with amplification, not just a timeout.
4. Worse, `linuxInstaller.capture()` then records the stale sentinel file as the "original", so disconnect restores broken DNS.

Two cheap defenses, both worth doing:

- Filter the server's own bind IP out of captured upstreams. `buildResolvConf` already does exactly this for secondary nameservers (`if ns == bindIP { continue }`), so the forwarder is asymmetric today.
- Make `capture()` refuse to treat a sentinel-bearing file as the original (treat it as "no usable original" and log loudly).

There's also a milder in-session variant: on a tunnel-proxy restart where the outgoing service's resolv.conf restore failed, `New` runs `captureUpstreams` before `newInstaller`'s recovery, so the forwarder can capture the previous (dead) bind address as first upstream and eat the 2s per-server timeout on every forwarded query for the rest of the session. The self-filter above covers most of it.

### 2. macOS: "writing more than 10 resolver files" warns every refresh cycle, forever

`darwinInstaller.install()` emits the `len(desired) > 10` warning unconditionally, and the refresh loop calls `install()` every 15s. Observed live with 12 zones on the demo cluster: ~4 warnings/minute into `root-manager.log` for the lifetime of the session (thousands of lines/day). Fix: gate the warning on the desired set changing (there's already a `changed` flag in that function) or on first crossing the threshold.

---

Not a bug, but noticed live and worth a wording tweak: `local status` reports "147 cluster names resolvable via local DNS" while `local hosts` lists 49 rows. Both are correct (147 is exactly 49 x 3: each FQDN plus its two synthesized short forms), but a user comparing them will wonder where 98 names went. Something like "147 names (49 hosts)" in the status line, or reporting the host count there, would preempt the confusion.

PR #364, opened 2026-08-24
fix(print): handle carriage returns in --attach text output

jchak7:
While reading the --attach output code I noticed the text formatter can leak raw carriage returns into terminal output.

In formatText, the log message is trimmed with strings.TrimRight(e.Msg, "\n"), which drops a trailing \n but leaves a trailing \r. And quoteIfNeeded forces quoting on " \t\n\"=" but not \r. So a log line ending in \r\n prints as msg=<text> followed by a bare \r, and an embedded \r inside a message prints raw too. On a terminal that \r sends the cursor back to the start of the line and the next write overwrites it. CRLF endings are common (Windows apps, plenty of containers), so this turns up in real logs.

The fix is small:
- trim "\r\n" instead of "\n"
- add \r to the set that forces quoting, so an embedded \r is escaped as \r instead of emitted raw

JSON mode already handled this, since json.Marshal escapes control characters, so this only affects the text output.

This package had no tests, so I added the first ones. They cover quoteIfNeeded across the special characters and the log path for trailing CRLF, embedded CR, and a plain message.

    go test ./internal/print/
    ok  github.com/signadot/cli/internal/print

I checked each assertion pins the fix: reverting the trim fails only TestFormatTextLogTrimsTrailingCRLF, and reverting the quoteIfNeeded change fails only the carriage-return cases. gofmt is clean.

### DIFF
--- a/internal/print/attach.go
+++ b/internal/print/attach.go
@@ -78,7 +78,7 @@ func formatText(e AttachEvent) string {
 		if e.Stream != "" {
 			fmt.Fprintf(&b, " stream=%s", e.Stream)
 		}
-		fmt.Fprintf(&b, " msg=%s", quoteIfNeeded(strings.TrimRight(e.Msg, "\n")))
+		fmt.Fprintf(&b, " msg=%s", quoteIfNeeded(strings.TrimRight(e.Msg, "\r\n")))
 	case "output":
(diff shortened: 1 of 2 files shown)


foxish (approved):
LGTM. Thanks for the fix and for adding the first tests to this package, much appreciated!

